PR #18, opened 2017-03-09
CC-497: Add timestamp based partitioners.

kkonstantine:
/cc @confluentinc/connect

### DIFF
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
@@ -95,7 +95,7 @@ public class PartitionerConfig extends AbstractConfig implements ComposableConfi
   public static final String TIMESTAMP_FIELD_NAME_CONFIG = "timestamp.field";
   public static final String TIMESTAMP_FIELD_NAME_DOC =
       "The record field to be used as timestamp by the timestamp extractor.";
-  public static final String TIMESTAMP_FIELD_NAME_DEFAULT = "Timestamp";
+  public static final String TIMESTAMP_FIELD_NAME_DEFAULT = "timestamp";
   public static final String TIMESTAMP_FIELD_NAME_DISPLAY = "Record Field for Timestamp Extractor";
 
   // CHECKSTYLE:OFF
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
@@ -43,6 +43,11 @@ public class PartitionerConfig extends AbstractConfig implements ComposableConfi
   public static final Class<?> PARTITIONER_CLASS_DEFAULT = DefaultPartitioner.class;
   public static final String PARTITIONER_CLASS_DISPLAY = "Partitioner Class";
 
+  public static final String SCHEMA_GENERATOR_CLASS_CONFIG = "schema.generator.class";
+  public static final String SCHEMA_GENERATOR_CLASS_DOC =
+      "The schema generator to use with partitioners.";
+  public static final String SCHEMA_GENERATOR_CLASS_DISPLAY = "Schema Generator Class";
+
   public static final String PARTITION_FIELD_NAME_CONFIG = "partition.field.name";
   public static final String PARTITION_FIELD_NAME_DOC =
       "The name of the partitioning field when FieldPartitioner is used.";
@@ -59,10 +64,10 @@ public class PartitionerConfig extends AbstractConfig implements ComposableConfi
   public static final String PATH_FORMAT_CONFIG = "path.format";
   public static final String PATH_FORMAT_DOC =
       "This configuration is used to set the format of the data directories when partitioning with "
-      + "``TimeBasedPartitioner``. The format set in this configuration converts the Unix timestamp "
-      + "to proper directories strings. For example, if you set "
-      + "``path.format='year'=YYYY/'month'=MM/'day'=dd/'hour'=HH/``, the data directories will have"
-      + " the format ``/year=2015/month=12/day=07/hour=15``.";
+      + "``TimeBasedPartitioner``. The format set in this configuration converts the Unix timestamp"
+      + " to proper directories strings. For example, if you set "
+      + "``path.format='year'=YYYY/'month'=MM/'day'=dd/'hour'=HH``, the data directories will have"
+      + " the format ``/year=2015/month=12/day=07/hour=15/``.";
   public static final String PATH_FORMAT_DEFAULT = "";
   public static final String PATH_FORMAT_DISPLAY = "Path Format";
 
@@ -78,10 +83,18 @@ public class PartitionerConfig extends AbstractConfig implements ComposableConfi
   public static final String TIMEZONE_DEFAULT = "";
   public static final String TIMEZONE_DISPLAY = "Timezone";
 
-  // Schema group
-  public static final String SCHEMA_GENERATOR_CLASS_CONFIG = "schema.generator.class";
-  public static final String SCHEMA_GENERATOR_CLASS_DOC = "The schema generator to use with partitioners.";
-  public static final String SCHEMA_GENERATOR_CLASS_DISPLAY = "Schema Generator Class";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_CONFIG = "timestamp.extractor";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_DOC =
+      "The extractor used to get timestamp for records when partitioning with "
+          + "``TimeBasedPartitioner``.";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_DEFAULT = "Wallclock";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_DISPLAY = "Timestamp Extractor";
+
+  public static final String TIMESTAMP_FIELD_NAME_CONFIG = "timestamp.field";
+  public static final String TIMESTAMP_FIELD_NAME_DOC =
+      "The record field to be used as timestamp by the timestamp extractor.";
+  public static final String TIMESTAMP_FIELD_NAME_DEFAULT = "Timestamp";
+  public static final String TIMESTAMP_FIELD_NAME_DISPLAY = "Record Field for Timestamp Extractor";
 
   // CHECKSTYLE:OFF
   public static final ConfigDef.Recommender partitionerClassDependentsRecommender =
@@ -166,6 +179,25 @@ public class PartitionerConfig extends AbstractConfig implements ComposableConfi
           Width.LONG,
           TIMEZONE_DISPLAY);
 
+      CONFIG_DEF.define(TIMESTAMP_EXTRACTOR_CLASS_CONFIG,
+          Type.STRING,
+          TIMESTAMP_EXTRACTOR_CLASS_DEFAULT,
+          Importance.MEDIUM,
+          TIMESTAMP_EXTRACTOR_CLASS_DOC,
+          group,
+          ++orderInGroup,
+          Width.LONG,
+          TIMESTAMP_EXTRACTOR_CLASS_DISPLAY);
+
+      CONFIG_DEF.define(TIMESTAMP_FIELD_NAME_CONFIG,
+          Type.STRING,
+          TIMESTAMP_FIELD_NAME_DEFAULT,
+          Importance.MEDIUM,
+          TIMESTAMP_FIELD_NAME_DOC,
+          group,
+          ++orderInGroup,
+          Width.LONG,
+          TIMESTAMP_FIELD_NAME_DISPLAY);
     }
   }
 
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
@@ -84,9 +84,11 @@ public class PartitionerConfig extends AbstractConfig implements ComposableConfi
   public static final String TIMEZONE_DISPLAY = "Timezone";
 
   public static final String TIMESTAMP_EXTRACTOR_CLASS_CONFIG = "timestamp.extractor";
-  public static final String TIMESTAMP_EXTRACTOR_CLASS_DOC =
-      "The extractor used to get timestamp for records when partitioning with "
-          + "``TimeBasedPartitioner``.";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_DOC = "The extractor that gets the "
+      + "timestamp for records when partitioning with ``TimeBasedPartitioner``. It can be set to "
+      + "``Wallclock``, ``Record`` or ``RecordField`` in order to use one of the built-in "
+      + "timestamp extractors or be given the fully-qualified class name of a user-defined class "
+      + "that extends the ``TimestampExtractor`` interface.";
   public static final String TIMESTAMP_EXTRACTOR_CLASS_DEFAULT = "Wallclock";
   public static final String TIMESTAMP_EXTRACTOR_CLASS_DISPLAY = "Timestamp Extractor";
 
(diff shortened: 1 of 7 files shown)


kkonstantine:
This PR bundles three types of time-based partitioners: 
- Wall-clock-based (still the default). Depending on the store, it might not guarantee exactly-once semantics.
- Record timestamp-based. Buckets kafka records according to their record timestamp. Requires timestamps to be enabled for records in Kafka. Guarantees exactly-once semantics. 
- Record field-based. Buckets kafka records according to a user-defined record field that represents a timestamp. Guarantees exactly-once semantics.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -103,8 +137,87 @@ public String encodePartition(SinkRecord sinkRecord) {
...
+          case INT8:
+          case INT16:
+          case INT32:
+          case INT64:
+            Number timestamp = (Number) timestampValue;
+            return timestamp.longValue();
+          case STRING:
+            // Should we support parsing from string?
+          case BYTES:
+            // Do we support serialized timestamps?
```
kkonstantine: @ewencp @gwenshap Will be writing tests for record-field user-defined timestamps and will try to be as liberal as I can regarding what is accepted as a user-defined timestamp column, but suggestions are more than welcome.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
```diff
@@ -78,10 +82,17 @@
...
   public static final String TIMEZONE_DISPLAY = "Timezone";
 
-  // Schema group
-  public static final String SCHEMA_GENERATOR_CLASS_CONFIG = "schema.generator.class";
-  public static final String SCHEMA_GENERATOR_CLASS_DOC = "The schema generator to use with partitioners.";
-  public static final String SCHEMA_GENERATOR_CLASS_DISPLAY = "Schema Generator Class";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_CONFIG = "timestamp.extractor";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_DOC =
+      "The extractor used to get timestamp for records when partitioning with ``TimeBasedPartitioner``.";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_DEFAULT = "Wallclock";
```
ewencp: If there are a fixed set, validator?

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -17,31 +17,47 @@
...
                       Map<String, Object> config) {
     delim = (String) config.get(StorageCommonConfig.DIRECTORY_DELIM_CONFIG);
     this.partitionDurationMs = partitionDurationMs;
+    this.pathFormat = pathFormat;
     this.formatter = getDateTimeFormatter(pathFormat, timeZone).withLocale(locale);
     try {
       partitionFields = newSchemaGenerator(config).newPartitionFields(pathFormat);
+      timestampExtractor = newTimestampExtractor(
+          (String) config.get(PartitionerConfig.TIMESTAMP_EXTRACTOR_CLASS_CONFIG));
+      timestampExtractor.configure(config);
```
ewencp: Coming back to the "it's not a class" comment above, if we have the same structure, is there any reason *not* to just make it a pluggable class?

The reason I ask is that even fields may not be ideal, e.g. if your timestamp is embedded deeper in the message, e.g. 
```
{
  "header": {
    "timestamp": 100,
    "dc": "dc1",
  },
  "payload": { ... }
}
```
 Of course with an SMT you can always pull that up to a top-level field and building too much flexibility into extracting the timestamp might be undesirable, but then you could also just use  an SMT to set the record's timestamp (with more flexibility, e.g. using dotted notation) and then only offer wallclock or record?

partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
ewencp: Ok, I see now that Wallclock, Record, and RecordField are special cases. Add to the doc string? Otherwise I wouldn't have even thought to try the other two.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -103,8 +137,87 @@ public String encodePartition(SinkRecord sinkRecord) {
...
+    public Long extract(ConnectRecord<?> record) {
+      Object value = record.value();
+      Schema valueSchema = record.valueSchema();
+      if (value instanceof Struct) {
+        Struct struct = (Struct) value;
+        Object timestampValue = struct.get(fieldName);
+        Type type = valueSchema.field(fieldName).schema().type();
+        switch (type) {
+          case INT8:
+          case INT16:
```
ewencp: I don't know that I'd even bother with `INT8` and `INT16`. They aren't large enough to reasonably represent a timestamp...

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -103,8 +137,87 @@ public String encodePartition(SinkRecord sinkRecord) {
...
+          case INT32:
+          case INT64:
+            Number timestamp = (Number) timestampValue;
+            return timestamp.longValue();
+          case STRING:
+            // Should we support parsing from string?
```
ewencp: https://en.wikipedia.org/wiki/ISO_8601 is what I was referring to. We already use joda-time, right? It should have built-in support for this.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
ewencp: Serialized according to what format? I think we can likely just ignore `BYTES`. The only case I can think of that makes sense here is if it's actually a `BigDecimal`.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -103,8 +137,87 @@ public String encodePartition(SinkRecord sinkRecord) {
...
+          case BYTES:
+            // Do we support serialized timestamps?
+          default:
+            log.error("Type {} is not supported as a user-defined record timestamp field.", type.getName());
+            throw new PartitionException("Error extracting timestamp from record field: " + fieldName);
+        }
+      } else {
+        log.error("Value is not Struct type.");
```
ewencp: Might want to support `Map`s as well for when the data is schemaless.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -103,8 +137,87 @@ public String encodePartition(SinkRecord sinkRecord) {
...
+
+    @Override
+    public Long extract(ConnectRecord<?> record) {
+      Object value = record.value();
+      Schema valueSchema = record.valueSchema();
+      if (value instanceof Struct) {
+        Struct struct = (Struct) value;
+        Object timestampValue = struct.get(fieldName);
+        Type type = valueSchema.field(fieldName).schema().type();
+        switch (type) {
```
ewencp: To handle logical types here as well, you can just check if the names match the pre-defined names for logical types in Connect. For example, in AvroConverter we use this to check if there is a special process for converting those types: https://github.com/confluentinc/schema-registry/blob/master/avro-converter/src/main/java/io/confluent/connect/avro/AvroData.java#L1207-L1212 In this case you could use it to first check for special logical types and handle them (e.g. timestamps in Connect are Java `Date`s iirc), then if it isn't a special logical type work off of the normal base type.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
```diff
@@ -61,8 +65,8 @@
       "This configuration is used to set the format of the data directories when partitioning with "
       + "``TimeBasedPartitioner``. The format set in this configuration converts the Unix timestamp "
       + "to proper directories strings. For example, if you set "
-      + "``path.format='year'=YYYY/'month'=MM/'day'=dd/'hour'=HH/``, the data directories will have"
-      + " the format ``/year=2015/month=12/day=07/hour=15``.";
+      + "``path.format='year'=YYYY/'month'=MM/'day'=dd/'hour'=HH``, the data directories will have"
```
kkonstantine: I wasn't feeling very strong about this being added to the doc. However it's there to denote that it's a directory path. 

Re: this property I added specific checks to make sure that `//` is avoided in S3.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
kkonstantine: I wrote it like that as a tradeoff between having to write the whole canonical name for this class and having to scan the classpath to match the simple name. 

I think it's good enough for a property that will mostly have one of these three values. Definitely the options need to be listed in the doc string.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
```diff
@@ -78,10 +82,17 @@
...
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_CONFIG = "timestamp.extractor";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_DOC =
+      "The extractor used to get timestamp for records when partitioning with ``TimeBasedPartitioner``.";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_DEFAULT = "Wallclock";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_DISPLAY = "Timestamp Extractor";
+
+  public static final String TIMESTAMP_FIELD_NAME_CONFIG = "timestamp.field";
+  public static final String TIMESTAMP_FIELD_NAME_DOC =
+      "The record field to be used as timestamp by the timestamp extractor.";
+  public static final String TIMESTAMP_FIELD_NAME_DEFAULT = "Timestamp";
```
kkonstantine: Again, added as a reasonable tradeoff between "no default" which would make this property unnecessarily mandatory and "empty string" which is practically a useless default. 

I'm fine with any reasonable name. That's something anyone who uses custom timestamp fields will probably change anyways.  I'd go with `timestamp`

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
kkonstantine: I think the confusion regarding whether this property can accept a custom class has been addressed in the comment re: the config above.

The user will always give a string. Either one of `Wallclock`, `Record`, or `RecordField` or the canonical class name of a custom class, such as `io.mycomp.mypackage.NestedRecordField`(of course they can still give the canonical name of the predefined class such as `io.confluent.connect.storage.partitioner.WallclockTimestampExtractor` but I don't see any advantage in doing so)

Now whether we treat this property as `Type.STRING` (as we do right now), or `Type.CLASS`, that's something internal. It was easier to treat it as a string for now. If there's a good way with a verifier/recommender to treat it as `Type.CLASS` we can change it later. Let me know if you agree.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
kkonstantine: Well yes, of course, the question is more about which format type(s) we expect in order to decide which joda-time formatter to use. I feel we should go with just one for now, such as [ISODateTimeFormat.dateTime](http://www.joda.org/joda-time/apidocs/org/joda/time/format/ISODateTimeFormat.html#dateTime--) which parses date,time up to millis and use the timezone that's supplied by the user.

kkonstantine:
Thanks for your review @ewencp. I replied to all the comments. Some might need more discussion. Also tests for the moment reside in `storage-cloud`, so that's were they'll be added for the user-defined timestamp field.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -103,8 +139,102 @@ public String encodePartition(SinkRecord sinkRecord) {
...
+  }
+
+  public static class RecordFieldTimestampExtractor implements TimestampExtractor {
+    private String fieldName;
+    private DateTimeFormatter dateTime;
+
+    @Override
+    public void configure(Map<String, Object> config) {
+      fieldName = (String) config.get(PartitionerConfig.TIMESTAMP_FIELD_NAME_CONFIG);
+      dateTime = ISODateTimeFormat.dateTime();
```
kkonstantine: I assumed the user-defined timestamp field won't need time zone conversion. But will revisit during tests.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
```diff
@@ -78,10 +82,20 @@
...
-  public static final String SCHEMA_GENERATOR_CLASS_CONFIG = "schema.generator.class";
-  public static final String SCHEMA_GENERATOR_CLASS_DOC = "The schema generator to use with partitioners.";
-  public static final String SCHEMA_GENERATOR_CLASS_DISPLAY = "Schema Generator Class";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_CONFIG = "timestamp.extractor";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_DOC =
+      "The extractor that gets the timestamp for records when partitioning with ``TimeBasedPartitioner``. It can be " +
+          "set to ``Wallclock``, ``Record`` or ``RecordField`` in order to use one of the built-in timestamp " +
+          "extractors or be given the fully-qualified class name of a user-defined class that extends the" +
+          "``TimestampExtractor`` interface.";
+  public static final String TIMESTAMP_EXTRACTOR_CLASS_DEFAULT = "Wallclock";
```
ewencp: Is this ok as a default given https://github.com/confluentinc/kafka-connect-storage-cloud/blob/master/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L193?

kkonstantine: No, the changes here need to happen in lock-step with the changes in the equivalent PR in storage-cloud: https://github.com/confluentinc/kafka-connect-storage-cloud/pull/19

The section in the link you mention above, becomes:
https://github.com/kkonstantine/kafka-connect-storage-cloud/blob/29f3da40ebea2ef4f1eb3a0f733c9867e7f1148f/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L164

kurtostfeld:
I just submitted a pull request which looks like it is redundant with this one. Two suggestions that are missing from this pull request: PARTITION_DURATION_MS should be optional. If it's not specified, not extra rounding should be done. The date format string will naturally partition timestamp values. Typically, that's all that is desired and requiring an extra setting is confusing. Secondly, the `fieldName` variable in the `FieldPartitioner` class was obviously not supposed to be static. That is a bug that should be fixed.

kkonstantine:
Thanks @kurtostfeld for your comments. Good catch, I fixed `fieldName`. With `PARTITION_DURATION_MS` I agree there is some redundancy, but I'd prefer to address in a later PR, since this will be used by both S3 and HDFS connectors.

kkonstantine:
Thanks @ewencp for your review!

kkonstantine:
Merging after rebasing on top of master to get the checkstyle changes and back on top of 3.2.x to have it available at 3.2.x releases.

PR #19, opened 2017-03-27
move to using io.confluent:common for deps

norwood:
first part of this for connect stuff. will follow up with rest of connect repos if we are happy with this.


norwood:
dependent on https://github.com/confluentinc/common/pull/49

pom.xml
```diff
@@ -82,30 +83,55 @@
...
-              </plugin>
-            </plugins>
-        </pluginManagement>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>2.5.1</version>
                 <inherited>true</inherited>
                 <configuration>
```
ewencp: Should the config for this be going all the way up in common's pluginManagement? I realize most of the contents here are overridden compiler arguments, but the 1.7 source/target in common will be handy when we want to change it to 1.8.

core/src/main/java/io/confluent/connect/storage/Storage.java
```diff
@@ -53,30 +56,34 @@
...
+   * For instance successfully closing the stream after
+   * writing all the data. Additionally, the effect of this
+   * operation might not be immediately visible to subsequent
    * operations.
    *
    * @param path the path of the object to be created.
    * @param conf storage configuration.
-   * @param overwrite whether to override an existing object with the same path (optional operation).
+   * @param overwrite whether to override an existing object with
+   *                  the same path (optional operation).
```
kkonstantine: I now see that the checkstyle property that complains about alignment here is: `JavadocTagContinuationIndentation`
How strong do we feel about this property? Seems to create problems with intellij: https://github.com/checkstyle/checkstyle/issues/3258

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -100,10 +123,16 @@ public String encodePartition(SinkRecord sinkRecord) {
...
     try {
       generatorClass =
-          (Class<? extends SchemaGenerator<T>>) config.get(PartitionerConfig.SCHEMA_GENERATOR_CLASS_CONFIG);
+          (Class<? extends SchemaGenerator<T>>) config.get(
+              PartitionerConfig.SCHEMA_GENERATOR_CLASS_CONFIG
+          );
       return generatorClass.getConstructor(Map.class).newInstance(config);
-    } catch (ClassCastException | IllegalAccessException | InstantiationException | InvocationTargetException
-        | NoSuchMethodException e) {
+    } catch (ClassCastException
```
kkonstantine: Not sure about this one. Again not a method call. And `ClassCastException` is not on its own line and the indentation of the final `) {` is not exactly at the right level probably. 

Checkstyle is complaining at `OperatorWrap` but not if I write `catch` as follows: 

```java
tor(confClass, String.class);
      return constructor.newInstance(conf, url);
    } catch (NoSuchMethodException
        | InvocationTargetException
        | InstantiationException
        | IllegalAccessException e) {
      throw new ConnectException(e);
    }
```
it's not throwing an error. Your call.

core/src/main/java/io/confluent/connect/storage/StorageFactory.java
```diff
@@ -37,13 +38,20 @@
...
     try {
       Constructor<S> constructor =
           storageClass.getConstructor(confClass, String.class);
       return constructor.newInstance(conf, url);
-    } catch (NoSuchMethodException | InvocationTargetException | InstantiationException | IllegalAccessException e) {
+    } catch (NoSuchMethodException
+        | InvocationTargetException
+        | InstantiationException
+        | IllegalAccessException e
+    ) {
```
kkonstantine: Actually see comment about `catch` below. Checkstyle complains if written as above.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/DailyPartitioner.java
```diff
@@ -30,20 +30,26 @@
 
   @Override
   public void configure(Map<String, Object> config) {
-    long partitionDurationMs = TimeUnit.HOURS.toMillis(24);
+    final long partitionDurationMs = TimeUnit.HOURS.toMillis(24);
```
norwood: this is here because the variable is declared too far away from where it is actually used.

i guess the other option would be to move this variable down lower

pom.xml
norwood: they are. i can remove the 1.7 pieces.

maybe i should move the xlint stuff up there also?

core/src/main/java/io/confluent/connect/storage/Storage.java
norwood: @dguy @ewencp

core/src/main/java/io/confluent/connect/storage/tools/SchemaSourceTask.java
```diff
@@ -98,10 +99,16 @@ public void start(Map<String, String> props) {
       topic = props.get(TOPIC_CONFIG);
       maxNumMsgs = Long.parseLong(props.get(NUM_MSGS_CONFIG));
       multipleSchema = Boolean.parseBoolean(props.get(MULTIPLE_SCHEMA_CONFIG));
-      partitionCount = Integer.parseInt(props.containsKey(PARTITION_COUNT_CONFIG) ?
-                                        props.get(PARTITION_COUNT_CONFIG) : "1");
-      isOutputEnabled = Boolean.parseBoolean(props.containsKey(ENABLE_STDOUT_CONFIG) ?
-                                        props.get(ENABLE_STDOUT_CONFIG) : "true");
+      partitionCount = Integer.parseInt(
+          props.containsKey(PARTITION_COUNT_CONFIG)
```
norwood: i think i like it broken up more, else its tough to *know* via scanning that theres a ternary here.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/DailyPartitioner.java
kkonstantine: I see. Fair point. Closer is better then. Because there's nothing special to this variable compared to other ones in this method.

core/src/main/java/io/confluent/connect/storage/tools/SchemaSourceTask.java
kkonstantine: Fair enough. I won't object. Though, if you ask me, when I use ternary operator for a conditional assignment I go for brevity (ideally a one-liner) more than anything else. At least when one of the two terms is simple (null, 1, true, etc)

PR #21, opened 2017-04-13
Added TimeFieldPartitioner that allows partitioning on a date-time field.

kurtostfeld:
The main feature I want to add is the `TimeFieldPartitioner` class that allows partitioning based on a date-time field. Currently, `FieldPartitioner` doesn't support date-time fields, and will simply generate a runtime exception. This new class uses the exact same configuration settings and partition naming logic of `TimeBasedPartitioner` but instead of using wall time, it uses the time value from the specified field.

I'm also adding some related bug fixes:

- Major Bug fix to `FieldPartitioner` where `fieldName` was a global static. It should be instance based. This was obviously a mistake in the original code.
- Minor fix to `HourlyPartitioner`: uses configurable `delim` field to build the path format string rather than assuming `/`. This behavior is identical to existing behavior in `DailyPartitioner`.
- Minor fix to `DailyPartitioner`: Added `PARTITION_DURATION_MS` constant to be analogous to the exiting constant in the similar `HourlyPartitioner` class.
- Feature: Allow `PARTITION_DURATION_MS_CONFIG` to not be user configured. Usually, this isn't needed. The date-time string formatting will automatically truncate to natural bucket partitions. Many users won't need additional date-time bucket rounding logic.
- Very minor code reorg to minimize code duplication related to adding `TimeFieldPartitioner`.

### DIFF
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/DailyPartitioner.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/DailyPartitioner.java
@@ -16,39 +16,32 @@
 
 package io.confluent.connect.storage.partitioner;
 
+import io.confluent.connect.storage.common.SchemaGenerator;
 import org.apache.kafka.common.config.ConfigException;
-import org.joda.time.DateTimeZone;
 
-import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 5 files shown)


kurtostfeld:
[clabot:check]

kurtostfeld:
Another pull request supersedes this.

PR #23, opened 2017-05-24
HOTFIX: Add test schema and record builder for records with timestamp field

kkonstantine:


partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -235,7 +235,7 @@ public Long extract(ConnectRecord<?> record) {
         Schema valueSchema = record.valueSchema();
         Schema fieldSchema = valueSchema.field(fieldName).schema();
 
-        if (fieldSchema.name().equals(Timestamp.LOGICAL_NAME)) {
+        if (Timestamp.LOGICAL_NAME.equals(fieldSchema.name())) {
```
kkonstantine: This shouldn't happen (and thus the order shouldn't matter) if the struct is correct, but I think it's better to fail in the conversion rather than throwing a NPE here.

ewencp:
retest this please

kkonstantine:
Merging as is due to its urgency and will add more tests cases in the next iteration. Thanks!

ewencp:
retest this please

ewencp:
retest this please

PR #24, opened 2017-05-25
Convert - to _ in Hive table names

ewencp:


ewencp:
@kkonstantine This is needed for refactoring HDFS to use common because of `-` used in some test topic names (and was requested as a follow-up to replacing `.` with `_` in https://github.com/confluentinc/kafka-connect-hdfs/pull/164). I think this covers all the translations needed for valid Kafka topic names given the constraints on Hive table names.

PR #25, opened 2017-05-25
Include causes with ConfigExceptions caused by catching other exceptions

ewencp:


ewencp:
It is helpful to understand the root cause of these if they were caused by another exception, but `ConfigException` is weird in that none of the constructors accept a `cause`. This actually caused me quite a bit of pain because doing `throw new ConfigException("message", e);` *looks* ok, but actually it just sets the `value` field on `ConfigException` and no `cause` is ever reported or logged with the exception. This approach at least attaches the cause so you can find it in a debugger.

PR #26, opened 2017-05-25
Add constructors to SchemaGenerators to support no parameters and a Map config

ewencp:


ewencp:
@kkonstantine These are sort of internal, but it turns out that `DefaultSchemaGenerator` can be instantiated by `TimeBasedPartitioner` and it tries to use the wrong constructor. Simply supporting both constructors seems like the simplest solution.

PR #27, opened 2017-05-25
HOTFIX: Include a trailing delimiter when verifying data format for Hive

kkonstantine:


kkonstantine:
@ewen this is necessary fix for the verification to succeed with a correct path format now that a trailing delimiter is not required.

PR #28, opened 2017-05-25
HOTFIX: Add a no-op default for schema generator class config property

kkonstantine:


kkonstantine:
@ewencp will be overwritten (by instantiating a new config) by any top-level configs that need to offer defaults for compatibility, etc. (see related commit in hdfs connector refactoring PR)

ewencp:
@kkonstantine Any reason this shouldn't just be `null`? Putting an uninstantiable interface in there seems like it might be more confusing than an NPE would be.

kkonstantine:
From what I can recall, the main motivation should be not to add yet another exception (NPE in this case) among the exceptions that we catch when instantiating a schema generator, for instance here: https://github.com/confluentinc/kafka-connect-storage-common/blob/master/partitioner/src/main/java/io/confluent/connect/storage/partitioner/DefaultPartitioner.java#L68-L71 
and here: 
https://github.com/confluentinc/kafka-connect-storage-common/blob/master/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java#L169-L176

Not a big deal to go either way. What do you think @ewencp ?

ewencp:
Hmm, I thought I tested and didn't end up with an NPE. I don't feel too strongly either way, just feels weird filling in a default that isn't instantiable.

kkonstantine:
But if `config.get` on this property with default value returns `null`, we should have NPE, right? (on the highlighted parts above). Didn't try it yet, but I'll be very surprised if it doesn't.

kkonstantine:
Superseded by https://github.com/confluentinc/kafka-connect-storage-common/pull/31. Closing

PR #29, opened 2017-06-14
Timestamp information should be copied to projected SinkRecord

niksajakovljevic:
Projected `SinkRecord` was missing timestamp information from the original `SinkRecord`

### DIFF
--- a/core/src/main/java/io/confluent/connect/storage/schema/StorageSchemaCompatibility.java
+++ b/core/src/main/java/io/confluent/connect/storage/schema/StorageSchemaCompatibility.java
@@ -169,7 +169,9 @@ public enum StorageSchemaCompatibility implements SchemaCompatibility {
                projected.getKey(),
                currentValueSchema,
                projected.getValue(),
-               record.kafkaOffset()
+               record.kafkaOffset(),
+               record.timestamp(),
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/confluent/connect/storage/schema/StorageSchemaCompatibility.java
```diff
@@ -169,7 +169,9 @@ public SinkRecord project(
                projected.getKey(),
                currentValueSchema,
                projected.getValue(),
-               record.kafkaOffset()
+               record.kafkaOffset(),
+               record.timestamp(),
+               record.timestampType()
```
rhauch: Would it be better to instead call:

    record.newRecord(record.topic(), record.kafkaPartition(), currentKeySchema, projected.getKey(), currentValueSchema, projected.getValue(), record.timestamp())

This is the same, except the record's offset and typestamp type are retained automatically and thus don't have to be passed in.

kkonstantine:
Thanks @niksajakovljevic for the fix! It makes sense. I'll go ahead and merge it as is since we are close to the end of the release cycle and both versions are similar. @rhauch your comment is valid. Let's consolidate in one of the next iterations.

PR #30, opened 2017-06-23
multiple field partitioning

hansoloyi:
This allows partitioning across multiple fields

Changes: partition.field.name -> partition.field.names
Comma separated list of field names to partition across
```
partition.field.names=<field_1>,<field_2>,<field_3>
```
This does break Tests in kafka-connect-storage-cloud
Here is a Pull Request with additional tests and fixed tests for this change:

https://github.com/confluentinc/kafka-connect-storage-cloud/pull/59

### DIFF
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldPartitioner.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldPartitioner.java
@@ -17,12 +17,11 @@
 package io.confluent.connect.storage.partitioner;
 
 import org.apache.kafka.connect.data.Schema;
-import org.apache.kafka.connect.data.Schema.Type;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+import java.util.List;
 import java.util.Map;
 
 import io.confluent.connect.storage.common.StorageCommonConfig;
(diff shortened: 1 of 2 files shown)


hansoloyi:
@ConfluentCLABot [clabot:check]

partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldPartitioner.java
```diff
@@ -17,55 +17,76 @@
...
+    }
+
+    partitionFields = newSchemaGenerator(config).newPartitionFields(format.toString());
   }
 
   @Override
   public String encodePartition(SinkRecord sinkRecord) {
     Object value = sinkRecord.value();
     Schema valueSchema = sinkRecord.valueSchema();
+    StringBuilder partition = new StringBuilder();
```
kkonstantine: Instead of building the joined set of partition fields manually, I'd suggest adding all the fields in a `List` and then using `org.apache.kafka.common.utils.Utils.join` to return the final `String`.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldPartitioner.java
```diff
@@ -17,55 +17,76 @@
...
+            boolean booleanRecord = (boolean) partitionKey;
+            partition.append(fieldName + "=" + Boolean.toString(booleanRecord) + delim);
+            break;
+          default:
+            log.error("Type {} is not supported as a partition key.", type.getName());
+            throw new PartitionException("Error encoding partition.");
+        }
+      }
+
+      if (partition.length() > 0) {
```
kkonstantine: Again, this check won't be necessary if you use `org.apache.kafka.common.utils.Utils.join`

partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
```diff
@@ -48,11 +48,11 @@
       "The schema generator to use with partitioners.";
   public static final String SCHEMA_GENERATOR_CLASS_DISPLAY = "Schema Generator Class";
 
-  public static final String PARTITION_FIELD_NAME_CONFIG = "partition.field.name";
```
kkonstantine: I think we'd need to retain both properties for a while for the sake of backwards compatibility.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
```diff
@@ -48,11 +48,11 @@
...
+  public static final String PARTITION_FIELD_NAME_CONFIG = "partition.field.names";
   public static final String PARTITION_FIELD_NAME_DOC =
-      "The name of the partitioning field when FieldPartitioner is used.";
```
kkonstantine: Deprecation of a config property can be flagged only at the documentation level for the moment. For instance, we should extend the old config doc in a way that notifies of the new config and the deprecation of the old one: 

Example:
```java 
public static final String PARTITION_FIELD_NAME_DOC = 
    "The name of the partitioning field when FieldPartitioner is used"
    + " (This property is deprecated. Use " + PARTITION_FIELD_NAMES_CONFIG + " instead).";
```
(something like that).

kkonstantine: And add `PARTITION_FIELD_NAMES_CONFIG` (plural) of course.

NishanthShajahan: connect UI in CC only supports adding a single field now  . Wouldn't that need changes or  can we pass a  comma separated strings from CC (struct1.fieldname,struct2.fieldname) ?

NishanthShajahan:
connect UI in CC only supports adding a single field now . Wouldn't that need changes or can we pass a comma separated strings from CC (struct1.fieldname,struct2.fieldname) ?

kkonstantine:
@NishanthShajahan correct. The field is comma separated string representing a list now. See https://github.com/confluentinc/kafka-connect-storage-common/pull/57 for the actual implementation.

kkonstantine:
Superseded by: https://github.com/confluentinc/kafka-connect-storage-common/pull/57

PR #31, opened 2017-08-03
Remove schema.generator.class config and have Formats specify their own SchemaGenerator internally

ewencp:


ewencp:
@kkonstantine This is a blocker for the refactor of HDFS on common-storage. The last couple of Hive tests failing is due to using the wrong `schema.generator.class` being used. Including this config would be a backwards incompatible config change for existing users. But I don't really see much reason for this to be configurable anyway. Hive is pretty specific about how it defines the fields for the partition keys, so I don't think there's much room here for customization and I've yet to hear anyone request that feature.

In order to keep the cleaner dependencies between packages and avoid a required dependency on the hive module if you use `Partitioner`, I'm still instantiating the class dynamically in the `Partitioner` but doing it only when the `partitionFields` are requested and using class name constants to avoid a direct reference.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/DefaultPartitioner.java
```diff
@@ -53,19 +58,21 @@ public String generatePartitionedPath(String topic, String encodedPartition) {
...
   @SuppressWarnings("unchecked")
   public SchemaGenerator<T> newSchemaGenerator(Map<String, Object> config) {
     Class<? extends SchemaGenerator<T>> generatorClass = null;
     try {
-      generatorClass =
-          (Class<? extends SchemaGenerator<T>>) config.get(
-              PartitionerConfig.SCHEMA_GENERATOR_CLASS_CONFIG
-          );
+      generatorClass = (Class<? extends SchemaGenerator<T>>)
+          Class.forName(SCHEMA_GENERATOR_CLASS);
```
rhauch: Rather than use the class name constant here, what putting this line in a protected `Class<?
 extends SchemaGenerator<T> getSchemaGeneratorClass()` method? That way `TimeBasedPartitioner` ([here](https://github.com/confluentinc/kafka-connect-storage-common/pull/31/files#diff-35e4bfafabe9a14f5205e0b7bc968690R176)) could just override this method to return its preferred class rather than the more complex `newSchemaGenerator` method.

Not only does this allow `TimeBasedPartitioner` to inherit the more complex behavior, it also allows a custom subclass to do the same should they ever need it.

ewencp: Ack, updated accordingly

ewencp:
@kkonstantine Thoughts on this, especially given that you had #28 in progress?

hive/src/main/java/io/confluent/connect/storage/hive/schema/TimeBasedSchemaGenerator.java
```diff
@@ -42,13 +42,17 @@ public TimeBasedSchemaGenerator() {
   }
 
   public TimeBasedSchemaGenerator(Map<String, Object> config) {
-    this.config = config;
+    this.config = config == null ? new HashMap<String, Object>() : config;
```
kkonstantine: This is given in the constructor. Do you expect the contents of the Map to ever change beyond this point?
My guts say no, in which case `Collections.emptyMap()` would be preferable. But if there's even one case where it would make sense to add values to an initially empty `config` (for testing or not), I'd definitely keep it as is.

PR #35, opened 2017-09-15
Add missing modules to the dependencyManagement pom section so downstream projects will inherit the right version automatically.

ewencp:


ewencp:
This is needed for converting storage-cloud to common pom and not include unnecessary version info (and should be fixed in the HDFS connector's pom as well).

PR #36, opened 2017-09-28
CC-1153: Storage connectors can add recommendations for Class types

kkonstantine:


common/src/main/java/io/confluent/connect/storage/common/GenericClassRecommender.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.apache.kafka.common.config.ConfigDef;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+public class GenericClassRecommender implements ConfigDef.Recommender {
+  private final List<Object> validValues = new ArrayList<>();
```
rhauch: One option is to use a CopyOnWriteArrayList that is already threadsafe so you can remove the synchronization.

common/src/main/java/io/confluent/connect/storage/common/GenericClassRecommender.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public synchronized void addValidValues(Collection<Object> values) {
```
rhauch: If this is to be a generic _class_ recommender, should the `addValidValues` method accept a collection of `Class` instances? And would be good to genericize this class on the specific type of superclass, e.g., 

```
public class GenericClassRecommender<T>  implements ConfigDef.Recommender {
    ...
    public void addValidValues(Collection<Class<? extends T>> implementationClasses) {
        ...
    }
    ...
}
```

Might also be useful to add a varargs method:

```
    public void addValidValues(Class<? extends T>... implementationClasses) {
        ...
    }
```

common/src/main/java/io/confluent/connect/storage/common/StorageCommonConfig.java
```diff
@@ -53,6 +53,8 @@
   public static final String FILE_DELIM_DISPLAY = "File Delimiter";
 
   protected static final ConfigDef CONFIG_DEF = new ConfigDef();
+  public static final GenericClassRecommender STORAGE_CLASS_RECOMMENDER =
+      new GenericClassRecommender();
```
rhauch: The benefit of having the GenericClassRecommender use generics is that this could instead be of type `GenericClassRecommender<Storage<?,?>`, which seems like it'd be safer, if it could work.

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
```diff
@@ -93,6 +94,8 @@
   public static final String SCHEMA_CACHE_SIZE_DISPLAY = "Schema Cache Size";
 
   protected static final ConfigDef CONFIG_DEF = new ConfigDef();
+  public static final GenericClassRecommender FORMAT_CLASS_RECOMMENDER =
+      new GenericClassRecommender();
```
rhauch: Again, would `GenericClassRecommender<Format>`be a bit safer?

common/src/main/java/io/confluent/connect/storage/common/GenericClassRecommender.java
kkonstantine: I thought I'd change it to SortedSet instead, but since it should rather be just a `Set` I changed it to `CopyOnWriteArraySet`. Thanks

common/src/main/java/io/confluent/connect/storage/common/GenericClassRecommender.java
kkonstantine: I didn't bother with generics because the base method `validValues` is not generic anyways.

kkonstantine: Oh, yeah, I remembered also the real reason. There's an unfortunate layout of classes and interfaces in packages. `Storage` is not visible from common module (it's declare in core). 

I feel that the current change does not justify moving classes around modules at the moment.

common/src/main/java/io/confluent/connect/storage/common/StorageCommonConfig.java
kkonstantine: I agree. But see above restriction re: modules

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
kkonstantine: I agree. But see above restriction re: modules

common/src/main/java/io/confluent/connect/storage/common/GenericClassRecommender.java
rhauch: Yeah, that's good, too.

common/src/main/java/io/confluent/connect/storage/common/StorageCommonConfig.java
rhauch: What restriction?

rhauch: Ah, I see the earlier comment. Had to expand the outdated code to see it.

common/src/main/java/io/confluent/connect/storage/common/GenericClassRecommender.java
rhauch: I think it's really a shame that we can't override ConfigDef configurations, since that would make this so much easier. I'm fine with this, but how about logging an issue to enable redefining configurations?

common/src/main/java/io/confluent/connect/storage/common/GenericClassRecommender.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+
+public class GenericClassRecommender implements ConfigDef.Recommender {
+  // We need a set to avoid duplicates
+  private final Set<Object> validValues = new CopyOnWriteArraySet<>();
+
+  public void addValidValues(Collection<Object> values) {
+    validValues.addAll(values);
+  }
```
rhauch: I still think it'd be better to add an overloaded method that takes a vararg rather than a collection, as that would make it easier to do [this](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/100/files#diff-76b1431fba87d80cee884c1e9ae4ac20R171).

common/src/main/java/io/confluent/connect/storage/common/StorageCommonConfig.java
kkonstantine: Yeah, it was gone because of the `synchronized` keyword.

common/src/main/java/io/confluent/connect/storage/common/GenericClassRecommender.java
kkonstantine: Yeah, I've come to the opposite conclusion from you on that. I've listed in the past the reasons why `List` is more versatile than varargs, but I'll have to think about it for bit to give you all the pros and cons. There are one or two cases that `List` can be used more easily than an array. Unless there are very strict performance reasons I'm inclined to use collections vs arrays. Especially with Java 8, varargs have even fewer reasons to be used.

kkonstantine: In most cases though the two options are synonymous.

rhauch: Yes, I agree that in many cases they are similar. However, the one place where varargs is superior (even over Java 8 lambdas) is when you want to pass a list of literal values. And that's what is going to be the case in all connectors. For example, [this](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/100/files#diff-76b1431fba87d80cee884c1e9ae4ac20R170):

```
    StorageCommonConfig.STORAGE_CLASS_RECOMMENDER.addValidValues(
        Arrays.<Object>asList(S3Storage.class)
    );
```
turns into:
```
    StorageCommonConfig.STORAGE_CLASS_RECOMMENDER.addValidValues(
        S3Storage.class
    );
```

And [this](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/100/files#diff-76b1431fba87d80cee884c1e9ae4ac20R174):

```
   PartitionerConfig.PARTITIONER_CLASS_RECOMMENDER.addValidValues(
        Arrays.<Object>asList(
            DefaultPartitioner.class,
            HourlyPartitioner.class,
            DailyPartitioner.class,
            TimeBasedPartitioner.class,
            FieldPartitioner.class
        )
    );
```
turns into:
```
   PartitionerConfig.PARTITIONER_CLASS_RECOMMENDER.addValidValues(
            DefaultPartitioner.class,
            HourlyPartitioner.class,
            DailyPartitioner.class,
            TimeBasedPartitioner.class,
            FieldPartitioner.class
    );
```
Every place you want to use the existing method, you'll first have to create the list/set/collection. Why not just have a vararg method and create the set inside the method in one place? In fact, why even have a method that takes a collection?

kkonstantine: The initial discussion w.r.t generics makes a nice example of why I would prefer using a Collection vs a Vararg argument more often than not. 

In an effort to not get too much out of context here's the section from the generics FAQ that explains why varargs are not as flexible as collections due to the array creation they imply: [GenericsFAQ](http://www.angelikalanger.com/GenericsFAQ/FAQSections/ProgrammingIdioms.html#FAQ300) and which I'm sure you are aware of. 

On the same note, another reason to generally avoid excessive use of arrays (and therefore varargs) is that they don't implement `Iterable`. Thus what we might avoid initially by not using `Arrays.asList` we might have to pay down the road because using arrays does not allow us to use them as collections. 

Overall, I'd say I have a preference of using collections in general, and in the few cases where I need to initialize the collection with a few values (java 8 made a one-liner possible with streams and java 9 makes it even shorter with `List.of`) I use the idiom used above instead of `... new ArrayList<>... .add, .add, .add`. 

In any case, I think that what we discuss here is a matter of coding style.

rhauch: Correct, the proposed changes are certainly not wrong and it is a design choice.

common/src/main/java/io/confluent/connect/storage/common/GenericClassRecommender.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.kafka.common.config.ConfigDef;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+
+public class GenericClassRecommender implements ConfigDef.Recommender {
```
rhauch: I'm curious: why is this named GenericClassRecommender when it can hold any type of value?

kkonstantine: Of course, you're right. I thought I had mentioned but I didn't. Well, I didn't want to encourage a very broad usage because the pattern is not very solid (still, seems to be the best we can do right now without changing `ConfigDef` and `AbstractConfig`). As I mentioned, I started by targeting a generic implementation only to realize that some interfaces were not on the right modules. 

I'll change the name to reflect what you are pointing out, and hopefully it won't be used for more properties than absolutely necessary. Thanks

kkonstantine:
Thanks @rhauch ! Merging

PR #38, opened 2017-10-06
CC-1153: Isolate config instances and recommenders.

kkonstantine:


kkonstantine:
Make config def non-static to allow for config def isolation across the inheriting connectors.

common/src/main/java/io/confluent/connect/storage/common/StorageCommonConfig.java
```diff
@@ -52,17 +56,23 @@
   public static final String FILE_DELIM_DEFAULT = "+";
   public static final String FILE_DELIM_DISPLAY = "File Delimiter";
 
-  protected static final ConfigDef CONFIG_DEF = new ConfigDef();
-  public static final GenericRecommender STORAGE_CLASS_RECOMMENDER =
-      new GenericRecommender();
+  public static ConfigDef newConfigDef(Collection<GenericRecommender> recommenders) {
```
rhauch: First of all, why is this not just `Recommender`? Also, why not just pass the recommender as a single named method argument rather than in a collection? 

Same questions for the other two config classes.

kkonstantine: 1) That's valid now. I started thinking that values could be added at the base config, but there's no reason now. Will change to `ConfigDef.Recommender`
2) The fact that we only need one recommender per base config is purely incidental. I figured it's more extensible to account for multiple from the beginning. Not crazy about how this is written implying an order but feels good enough for now.

rhauch: These methods are static, so they can't be overloaded or overwritten. The use of `Collection<Recommender>` seems really complicated, and I'd prefer we just use a single simple `Recommender` parameter.

kkonstantine: These methods are not supposed to be overloaded. They are called from within the class or with the classname anyways. 

As I mentioned in the other repo too, with already a few class type properties and a few dozen properties overall, adding every recommender as a separate argument is not an obvious scalable solution. Map would be another option. 

Anyway, we'll cross this bridge when we come to it. Adding as simple argument at the moment.

common/src/main/java/io/confluent/connect/storage/common/StorageCommonConfig.java
```diff
@@ -52,17 +52,14 @@
   public static final String FILE_DELIM_DEFAULT = "+";
   public static final String FILE_DELIM_DISPLAY = "File Delimiter";
 
-  protected static final ConfigDef CONFIG_DEF = new ConfigDef();
-  public static final GenericRecommender STORAGE_CLASS_RECOMMENDER =
-      new GenericRecommender();
-
-  static {
+  public static ConfigDef newConfigDef(ConfigDef.Recommender recommender) {
```
rhauch: Wouldn't this be a bit more clear if named `storageClassRecommender`?

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
```diff
@@ -93,17 +92,14 @@
   public static final int SCHEMA_CACHE_SIZE_DEFAULT = 1000;
   public static final String SCHEMA_CACHE_SIZE_DISPLAY = "Schema Cache Size";
 
-  protected static final ConfigDef CONFIG_DEF = new ConfigDef();
-  public static final GenericRecommender FORMAT_CLASS_RECOMMENDER =
-      new GenericRecommender();
-
-  static {
+  public static ConfigDef newConfigDef(ConfigDef.Recommender recommender) {
```
rhauch: ... and this named `formatClassRecommender`?

partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
```diff
@@ -99,17 +98,14 @@
   public static final String TIMESTAMP_FIELD_NAME_DEFAULT = "timestamp";
   public static final String TIMESTAMP_FIELD_NAME_DISPLAY = "Record Field for Timestamp Extractor";
 
-  protected static final ConfigDef CONFIG_DEF = new ConfigDef();
-  public static final GenericRecommender PARTITIONER_CLASS_RECOMMENDER =
-      new GenericRecommender();
-
-  static {
+  public static ConfigDef newConfigDef(ConfigDef.Recommender recommender) {
```
rhauch: and finally this named `partitionerClassRecommender`?

kkonstantine:
@rhauch addressed the latest comments as well. Do you see anything else missing?

rhauch:
@kkonstantine looks great! You've got my approval.

PR #39, opened 2017-10-09
HOTFIX: Check simple class names in recommender.

kkonstantine:


kkonstantine:
As the removed methods shows, it used to be a comparison between a String property and a class name (FQN or simple). Now it should be comparison between two classes. But in order to preserve compatibility with the old partitioners in hdfs (e.g. `io.confluent.connect.hdfs.partitioner.DefaultPartitioner` we need to apply comparison with the simple name again).

kkonstantine:
Thanks @ewencp !

PR #42, opened 2017-10-27
Add adapter class for Avro Data Config

btnewton:
Adds ComposableAvroDataConfig to adapt AvroDataConfig to use `org.apache.kafka.common.config.AbstractConfig` for use in Connectors. 

Change requested in [PR 110](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/110) of kafka-connect-storage-cloud

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -26,6 +26,10 @@
         <version>3.3.1-SNAPSHOT</version>
     </parent>
 
+    <artifactId>kafka-connect-storage-common</artifactId>
+    <packaging>jar</packaging>
+    <name>kafka-connect-storage-common</name>
+
     <dependencies>
         <dependency>
             <groupId>io.confluent</groupId>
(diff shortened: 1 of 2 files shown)


kkonstantine:
Hi @btnewton , 

thank you for your patch here too. Now, I was giving the fix a thought a few days ago (before this task gets preempted by some high priority stuff) and I was thinking that since inheritance from `AvroDataConfig` doesn't really buy us anything here (because of the incompatibility between the two `AbstractConfig` classes), and we practically have to copy the configs to a different config class in storage-common, I'd suggest just adding those configs along with the property we were using from the start (`SCHEMAS_CACHE_SIZE_CONFIG`) and be done with it for the moment. 

If something changes to `AvroDataConfig` in the future, we may reconsider inheriting from this class then. 

This way, we won't have to replicate values between the old and the new cache properties and we won't distort the way these configs are displayed in other systems by rearranging their order and moving around their group.

kkonstantine:
I issued a PR according to the above, as promised, here: 
https://github.com/confluentinc/kafka-connect-storage-common/pull/45

kkonstantine:
This is superseded by https://github.com/confluentinc/kafka-connect-storage-common/pull/45, so I'm closing this one here.

PR #43, opened 2017-10-30
HOTFIX: Add recommender for schema generator classes only on 3.3.x.

kkonstantine:


rhauch:
@kkonstantine, are there any other recommenders that we should provide?

kkonstantine:
@rhauch not in this repo.

PR #45, opened 2017-10-31
HOTFIX: Consider all the properties in AvroDataConfig.

kkonstantine:


maurodoglio:
Hi @kkonstantine, awesome patch! From what I understand this will enable the enhanced schema feature in the hdfs sink. And that's exactly what I was looking for! 😃 🎉 
I see it has been a month since you open this pull request, do you know if this is going to be reviewed/merged anytime soon? (Sorry if I sound pushy, I don't mean to 😛 )

kkonstantine:
Hey @maurodoglio, yes it was blocked by the release process. 
It'll be merged soon.

kkonstantine:
Rebased. Looks ready, targeting the right repo.

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
```diff
@@ -92,6 +94,19 @@
   public static final int SCHEMA_CACHE_SIZE_DEFAULT = 1000;
   public static final String SCHEMA_CACHE_SIZE_DISPLAY = "Schema Cache Size";
 
+  public static final String ENHANCED_AVRO_SCHEMA_SUPPORT_CONFIG = "enhanced.avro.schema.support";
+  public static final boolean ENHANCED_AVRO_SCHEMA_SUPPORT_DEFAULT = false;
```
mageshn: is there a down side to not make this default behavior?

kkonstantine: Good point. I'm inclined to be conservative at this point, and enabled it by default in next regular version (`4.1.0`) instead of the bugfix version (`4.0.1`).

mageshn: 👍

kkonstantine:
Thanks! I agree in enabling it by default in `4.1.0` but not `4.0.1`.

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
rhauch: Would enabling this by default break any existing configurations or behavior?

kkonstantine: It shouldn't I believe, given that it adds support for unions and enums, but we'll have the chance to test even more.

ewencp: if it's purely bugfix, put it in the earlier branch as long as it isn't risky. also, if that's the case, would we really need a config?

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
```diff
@@ -214,4 +253,12 @@ public Object get(String key) {
   public StorageSinkConnectorConfig(ConfigDef configDef, Map<String, String> props) {
     super(configDef, props);
   }
+
+  public AvroDataConfig avroDataConfig() {
+    Map<String, Object> props = new HashMap<>();
+    props.put(SCHEMA_CACHE_SIZE_CONFIG, get(SCHEMA_CACHE_SIZE_CONFIG));
```
dcsobral: Actually, `SCHEMA_CACHE_SIZE_CONFIG` is a different key than `AvroDataConfig`'s `SCHEMAS_CACHE_SIZE_CONFIG`, so this doesn't create `AvroDataConfig` correctly. I've reported on #64, and provided a fix on #65.

PR #46, opened 2017-11-14
CC-1299: Create AVRO_CODEC ConfigKey

wicknicks:

### DIFF
--- a/common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java
+++ b/common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java
@@ -21,6 +21,7 @@ import org.apache.kafka.common.config.ConfigDef;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * A recommender which decides visibility based on the value of a parent config.
@@ -30,7 +31,18 @@ public class ParentValueRecommender implements ConfigDef.Recommender {
   protected String parentConfigName;
   protected Object parentConfigValue;
 
+  /**
+   * Construct a recommender with the name of parent config and its value for which {@link #visible}
+   * returns true.
+   *
+   * @param parentConfigName the name of the parent config
+   * @param parentConfigValue the value of the parent config for which this config
+   *        will be visible (can be null).
+   */
   public ParentValueRecommender(String parentConfigName, Object parentConfigValue) {
+    if (parentConfigName == null) {
+      throw new NullPointerException("parentConfigName cannot be null.");
+    }
     this.parentConfigName = parentConfigName;
     this.parentConfigValue = parentConfigValue;
   }
@@ -42,8 +54,6 @@ public class ParentValueRecommender implements ConfigDef.Recommender {
 
   @Override
   public boolean visible(String name, Map<String, Object> connectorConfigs) {
-    Object object = connectorConfigs.get(parentConfigName);
-    return object == parentConfigValue
-        || (object != null && object.equals(parentConfigValue));
+    return Objects.equals(connectorConfigs.get(parentConfigName), parentConfigValue);
   }
 }
--- a/common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java
+++ b/common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java
@@ -16,17 +16,15 @@
 
 package io.confluent.connect.storage.common;
 
-import org.apache.kafka.common.config.ConfigDef;
-
-import java.util.Collections;
-import java.util.List;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.Map;
 import java.util.Objects;
 
 /**
  * A recommender which decides visibility based on the value of a parent config.
  */
-public class ParentValueRecommender implements ConfigDef.Recommender {
+public class ParentValueRecommender extends GenericRecommender {
 
   protected String parentConfigName;
   protected Object parentConfigValue;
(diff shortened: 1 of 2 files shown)


common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java
```diff
@@ -0,0 +1,49 @@
...
+  @Override
+  public List<Object> validValues(String name, Map<String, Object> connectorConfigs) {
+    return Collections.emptyList();
+  }
+
+  @Override
+  public boolean visible(String name, Map<String, Object> connectorConfigs) {
+    Object object = connectorConfigs.get(parentConfigName);
+    return object == parentConfigValue
+        || (object != null && object.equals(parentConfigValue));
```
rhauch: What about using the following instead of these three lines:

    return Objects.equals(connectorConfigs.get(parentConfigName), parentConfigValue);

common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * A recommender which decides visibility based on the value of a parent config.
+ */
+public class ParentValueRecommender implements ConfigDef.Recommender {
+
+  protected String parentConfigName;
+  protected Object parentConfigValue;
+
+  public ParentValueRecommender(String parentConfigName, Object parentConfigValue) {
```
rhauch: Would be good to have JavaDoc for this constructor, which would outline whether `parentConfigValue` can be null (I think it can).

common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java
```diff
@@ -0,0 +1,49 @@
...
+    this.parentConfigName = parentConfigName;
+    this.parentConfigValue = parentConfigValue;
+  }
+
+  @Override
+  public List<Object> validValues(String name, Map<String, Object> connectorConfigs) {
+    return Collections.emptyList();
```
rhauch: Should the constructor also take a collection (or array for convenience) of recommended values that are valid when the parent value matches the specified value? Either way, we don't really want the list of valid values being empty, right?

Then, the different connectors could instantiate this recommender instance with:

```
private static final ParentValueRecommender AVRO_COMPRESSION_RECOMMENDER
      = new ParentValueRecommender(FORMAT_CLASS_CONFIG, 
                                   AvroFormat.class, 
                                   AVRO_SUPPORTED_CODECS);

File: common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java (hunk 1 above)
Comment (wicknicks) [reply common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java]: Yes, indeed. For some reason, I thought Objects was introduced in Java 8.

File: common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java (hunk 3 above)
Comment (wicknicks) [reply common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java]: I am a bit hesitant to add a static collection here. IIUC, the validValues(..) method returns a list which is computed based on the properties (parsed so far). Ideally, we would want to pass in a lambda function or method reference.

Comment (rhauch) [reply common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java]: IIRC, `validValues(...)` is called when the configuration property is considered to be visible, and so the valid values for a visible AvroCodec property would always be the `AVRO_SUPPORTED_CODECS`. But we could enforce that as well by having `validValues(...)` call `visible(...)` and return an empty list if not visible and the supplied values when visible.

But I understand you're point. Unfortunately we're still on Java 7 and can't use method references or lambdas, which makes this harder. I'd say it's okay for this class to not be quite that flexible now since we don't need it. If somebody needs it in the future, they might have to override `validValues` or modify it to take a `Supplier<List<Object>>`. In the meantime, let's just be sure that the JavaDoc for the constructor is clear on how the supplied valid values would be used.

File: common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+public class ParentValueRecommender implements ConfigDef.Recommender {
+
+  protected String parentConfigName;
+  protected Object parentConfigValue;
+
+  /**
+   * @param parentConfigName the name of the parent config
+   * @param parentConfigValue the value for which this config will be visibile (can be null).
+   */
+  public ParentValueRecommender(String parentConfigName, Object parentConfigValue) {
```
rhauch: Nit: Always add a title for the JavaDoc to make it complete.

common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java
wicknicks: Sounds good. Updated the Javadoc.

rhauch:
@wicknicks, please add the CC issue to the PR title and commit messages.

wicknicks:
retest this please

wicknicks:
retest this please

wicknicks:
retest this please

common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java
```diff
@@ -0,0 +1,69 @@
...
+   *
+   * @param parentConfigName the name of the parent config
+   * @param parentConfigValue the value of the parent config for which this config will be
+   *     visible (can be null).
+   * @param validValues a non-null collection of valid values for this ConfigKey (can be empty).
+   */
+  public ParentValueRecommender(String parentConfigName, Object parentConfigValue,
+                                Collection<Object> validValues) {
+    if (parentConfigName == null) {
+      throw new NullPointerException("parentConfigName cannot be null.");
```
kkonstantine:
```java
Objects.requireNonNull(parentConfigName, "parentConfigName cannot be null.");
```
is slightly more concise

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
```diff
@@ -120,6 +131,20 @@ public static ConfigDef newConfigDef(ConfigDef.Recommender formatClassRecommende
       );
 
       configDef.define(
+          AVRO_CODEC_CONFIG,
```
kkonstantine: Its importance is low and there are other configs with higher importance. Lets add it somewhere in the same group but further below

common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java
```diff
@@ -0,0 +1,69 @@
...
+   * Construct a recommender with the name of parent config and its value for which {@link #visible}
+   * returns true.
+   *
+   * @param parentConfigName the name of the parent config
+   * @param parentConfigValue the value of the parent config for which this config will be
+   *     visible (can be null).
+   * @param validValues a non-null array of valid values for this ConfigKey (can be empty).
+   */
+  public ParentValueRecommender(String parentConfigName, Object parentConfigValue,
+                                Object[] validValues) {
```
kkonstantine: old alignment style.

common/src/main/java/io/confluent/connect/storage/common/ParentValueRecommender.java
```diff
@@ -0,0 +1,69 @@
...
+   * Construct a recommender with the name of parent config and its value for which {@link #visible}
+   * returns true.
+   *
+   * @param parentConfigName the name of the parent config
+   * @param parentConfigValue the value of the parent config for which this config will be
+   *     visible (can be null).
+   * @param validValues a non-null collection of valid values for this ConfigKey (can be empty).
+   */
+  public ParentValueRecommender(String parentConfigName, Object parentConfigValue,
+                                Collection<Object> validValues) {
```
kkonstantine: old alignment style.

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
```diff
@@ -92,15 +92,26 @@
...
    *
    * @param formatClassRecommender A recommender for format classes shipping out-of-the-box with
    *     a connector. The recommender should not prevent additional custom classes from being
    *     added during runtime.
+   * @param  avroRecommender A recommender for avro compression codecs.
    * @return the newly created configuration definition.
    */
-  public static ConfigDef newConfigDef(ConfigDef.Recommender formatClassRecommender) {
+  public static ConfigDef newConfigDef(ConfigDef.Recommender formatClassRecommender,
+                                       ConfigDef.Recommender avroRecommender) {
```
kkonstantine: I missed to mention this. Sorry.

wicknicks: Thanks, @kkonstantine. Fixed this.

PR #49, opened 2017-11-28
CC-1333: Remove dependency on hive-exec and instead use hive-exec:core

wicknicks:
The hive-exec library bundles all its dependencies and this creates conflicts
with Connect's libraries (for example, avro-1.8.2 in Connect conflicts with the
avro(-1.7.7) classes bundled in hive-exec).

### DIFF
--- a/hive/pom.xml
+++ b/hive/pom.xml
@@ -30,6 +30,11 @@
     <packaging>jar</packaging>
     <name>kafka-connect-storage-hive</name>
 
+    <properties>
+        <commons.lang3.version>3.1</commons.lang3.version>
+        <kryo.version>2.22</kryo.version>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>io.confluent</groupId>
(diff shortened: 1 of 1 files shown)


rhauch:
@wicknicks how do these changes fix the Avro library conflict? Is Avro one of the libraries that is not included in the `core` classified hive-exec?

wicknicks:
@rhauch: yes, by using the `core` classifier, we pick up the hive-exec library which has no bundled dependencies (which means avro, protobuf, commons-lang3 etc are all removed). This should now contain hive classes (the org/apache/hadoop/hive/ql package).

rhauch:
This change means that the HDFS connector repo needs to change ([see this PR](https://github.com/confluentinc/kafka-connect-hdfs/pull/259)). Is there also an impact on the Storage Cloud repo?

wicknicks:
There were no changes needed on the storage cloud repo. It doesn't use hive, so its oblivious to the changes

rhauch:
@wicknicks, actually, I wonder if this should be made on an older branch. Remember that unlike AK, we issue PRs against the oldest branch we plan to update.

wicknicks:
Related PR: https://github.com/confluentinc/common/pull/82 and https://github.com/confluentinc/schema-registry/pull/677

wicknicks:
Closing this to open a new PR against 3.3.x

PR #50, opened 2017-11-29
CC-1333: Remove dependency on hive-exec and instead use hive-exec:core

wicknicks:
The hive-exec library bundles all its dependencies and this creates conflicts
with Connect's libraries (for example, avro-1.8.2 in Connect conflicts with the
avro(-1.7.7) classes bundled in hive-exec). We also need to add commons-lang3
and kryo and avro-mapred here, since they are not pulled in by hive-exec's pom
but required at runtime.

### DIFF
--- a/hive/pom.xml
+++ b/hive/pom.xml
@@ -30,6 +30,11 @@
     <packaging>jar</packaging>
     <name>kafka-connect-storage-hive</name>
 
+    <properties>
+        <commons.lang3.version>3.1</commons.lang3.version>
+        <kryo.version>2.22</kryo.version>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>io.confluent</groupId>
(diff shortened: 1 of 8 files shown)


rhauch:
@wicknicks should this issue and the commit titles have "CC-1333: " as a prefix?

wicknicks:
@rhauch, @kkonstantine: Could you please look at this PR. It is the last piece to remove the classpath conflicts in HDFS connector. The PR has changed from the last time where I need to bump up the version of the parent POM files, so it looks at the latest common jars.

rhauch:
@wicknicks this looks good, but the commits are missing the `CC-1333: ` prefix.

hive/pom.xml
```diff
@@ -23,13 +23,18 @@
...
+        <version>3.3.2-SNAPSHOT</version>
     </parent>
 
     <artifactId>kafka-connect-storage-hive</artifactId>
     <packaging>jar</packaging>
     <name>kafka-connect-storage-hive</name>
 
+    <properties>
+        <commons.lang3.version>3.1</commons.lang3.version>
+        <kryo.version>2.22</kryo.version>
```
tony810430: Hi @wicknicks 
Did you forget to add the property for `avro.version`?

wicknicks: It's defined here (https://github.com/confluentinc/common/blob/master/pom.xml#L48), so we can share it across all avro dependent repos.

tony810430: Yes, you are right. I missed it because I was depending on common v4.0.0 to build this.

PR #55, opened 2018-02-01
CC-1489: NPE on records with null timestamp.

kkonstantine:
Fixes https://github.com/confluentinc/kafka-connect-storage-cloud/issues/109

### DIFF
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
@@ -132,7 +132,15 @@ public class TimeBasedPartitioner<T> extends DefaultPartitioner<T> {
 
   @Override
   public String encodePartition(SinkRecord sinkRecord) {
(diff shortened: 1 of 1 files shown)


partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -132,7 +132,19 @@ public void configure(Map<String, Object> config) {
 
   @Override
   public String encodePartition(SinkRecord sinkRecord) {
-    long timestamp = timestampExtractor.extract(sinkRecord);
+    Long timestamp = timestampExtractor.extract(sinkRecord);
+    if (timestamp == null) {
+      String timestampExtractorClassName = timestampExtractor.getClass().getName();
+      log.error(
```
mageshn: nit : its the same message that's used for logger and the exception. May be we could create an error string and use it in both.

kkonstantine: Fixed

Issue #61, opened 2018-03-07
RecordFieldTimestampExtractor does not tolerate string timestamps without milliseconds


robobario:
Our application is emitting ISO8601 timestamps using java.util.Instant, and our kafka connect hdfs sink is barfing when it hits a timestamp without milliseconds:

`java.lang.IllegalArgumentException: Invalid format: "2015-04-02T01:00:00Z" is malformed at "Z`

jodatime's `ISODateTimeFormat.dateTime()` does not support millis but it has a `ISODateTimeFormat.dateTimeParser()` that supports both millis or no millis, [here's a discussion](https://github.com/JodaOrg/joda-time/issues/97) about it.

kkonstantine:
Feature was merged in `4.1.x` and later. Closing here, but feel free to reopen if the fix doesn't work for you.

PR #59, opened 2018-02-08
Add Partitioner unit tests

cricket007:
In addition to code added with #48 

Includes #63 

Fixes confluentinc/kafka-connect-hdfs#316, fixes https://github.com/confluentinc/kafka-connect-storage-cloud/issues/152

PR #62, opened 2018-03-07
Tolerate timestamp strings without milliseconds in RecordFieldTimesta…

robobario:
…mpExtractor

#61

### DIFF
--- a/core/src/main/java/io/confluent/connect/storage/StorageSinkTestBase.java
+++ b/core/src/main/java/io/confluent/connect/storage/StorageSinkTestBase.java
@@ -101,6 +101,10 @@ public class StorageSinkTestBase {
   }
 
   protected Schema createSchemaWithTimestampField() {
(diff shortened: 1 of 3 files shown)


robobario:
It looks like this would conflict with another open PR #59 adding unit tests to the partitioner.

cricket007:
> would conflict with another open PR

Feel free to PR into my branch, if you want.

kkonstantine:
Hey @robobario 
Nice find, thanks for the fix!
 
Somehow I read the github issue but I missed this PR. I found it after I had written the tests. Given we'd like to merge this fix soon, I added my unit tests on top of you branch here, and I'll be pushing to your branch so that you can keep your initial code change. Since I added tests for every timestamp type besides `string` I consolidated our changes within the unit test code. 

Also targeting `4.1.x` in order to include both tests and code (will backport the code fix alone further back in a separate PR). 

Thanks again!

PR #65, opened 2018-05-11
Fix #64 by using AvroDataConfig keys in the new map

### LINKED ISSUE #64: StorageSinkConnectorConfig#avroDataConfig() uses wrong AvroDataConfig keys
When creating an `AvroDataConfig` out of the values present in `StorageSinkConnectorConfig`, the method `avroDataConfig()` uses `StorageSinkConnectorConfig` as keys in the map, and, thus, end up not passing the schema cache config size, whose keys differ between them -- `schema.cache.size` on storage, `schemas.cache.config` on avro data.

dcsobral:
AvroDataConfig's SCHEMAS_CACHE_SIZE_CONFIG is not the same as StorageSinkConnectorConfig's SCHEMA_CACHE_SIZE_CONFIG, which leads to AvroDataConfig being created without the correct data for that configuration parameter.

### DIFF
--- a/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
+++ b/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
@@ -288,9 +288,10 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
 
   public AvroDataConfig avroDataConfig() {
     Map<String, Object> props = new HashMap<>();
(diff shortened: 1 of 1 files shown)


dcsobral:
I cannot reproduce this error locally, using the same command line. The console log has no clues about what went wrong, aside a generic "null pointer exception", and the build doesn't have any stored information about tests.

dcsobral:
Could not reproduce locally using same command line, and build doesn't contain information about test failures. Both this PR and one on schema-registry are failing with what seems to be the same error: an NPE when running the test goal, right on the first module after parent. The original build back in May passed, and a new build was run around June 7/8 (depending on repo) failing, and always failing from that time on.

PR #70, opened 2018-05-31
MINOR: Fix compilation error due to new method in SinkTaskContext

rhauch:
Recently added a new method in SinkTaskContext. We don't call it, so we can implement by throwing UnsupportedOperationException.

This only has to be merged to `master`.


rhauch:
The build fails but is part of a larger build pipeline that is somewhat circular. This builds locally with the latest from AK `trunk`, and this will then unblock the build pipeline. Merging despite the build failures.

PR #71, opened 2018-06-14
Add a method to cascade Hive alter changes

RussWinkler:
Add a new method that will allow the ability to cascade an alter table command in hive.


kkonstantine:
@RussWinkler I found an older PR adding this feature to the HDFS connector (https://github.com/confluentinc/kafka-connect-hdfs/pull/240). Do you mind opening a new PR that will use this method added here instead?

PR #63, opened 2018-04-19
DataUtils::getField returns field

yuryn:
In case of map argument DataUtils::getField does not return value, it cause raising exception as described here:

https://github.com/confluentinc/kafka-connect-hdfs/issues/316

PR #73, opened 2018-07-02
add new partitioner

shlomihassan:
add new type of patittioner

### DIFF
--- a/core/src/main/java/io/confluent/connect/storage/util/DataUtils.java
+++ b/core/src/main/java/io/confluent/connect/storage/util/DataUtils.java
@@ -33,6 +33,7 @@ public class DataUtils {
       if (field == null) {
         throw new DataException(String.format("Unable to find nested field '%s'", fieldName));
       }
+      return field;
     }
     throw new DataException(String.format(
           "Argument not a Struct or Map. Cannot get field '%s' from: %s",
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldAndTimeBasedPartitioner.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldAndTimeBasedPartitioner.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2017 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.confluent.connect.storage.partitioner;
+
+import io.confluent.connect.storage.common.StorageCommonConfig;
+import io.confluent.connect.storage.errors.PartitionException;
+import io.confluent.connect.storage.util.DataUtils;
+import org.apache.kafka.common.protocol.types.Field;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Schema.Type;
+import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Date;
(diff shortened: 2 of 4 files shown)


core/src/main/java/io/confluent/connect/storage/util/DataUtils.java
```diff
@@ -33,6 +33,13 @@ public static Object getField(Object structOrMap, String fieldName) {
       if (field == null) {
         throw new DataException(String.format("Unable to find nested field '%s'", fieldName));
       }
+      if (field == null) {
```
aviyehuda: It seems you accidentally copied this 3 lines

partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldAndTimeBasedPartitioner.java
```diff
@@ -0,0 +1,123 @@
...
+
+        Object partitionKey = struct.get(fieldName);
+        Type type = valueSchema.field(fieldName).schema().type();
+        switch (type) {
+          case INT8:
+          case INT16:
+          case INT32:
+          case INT64:
+            Number record = (Number) partitionKey;
+            builder.append(fieldName + "=" + record.toString());
```
aviyehuda: the toString() here is not needed.  when you do "" + someObject ,  it automatically does "" + someObject.toString()  .

partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldAndTimeBasedPartitioner.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        Object partitionKey = struct.get(fieldName);
+        Type type = valueSchema.field(fieldName).schema().type();
+        switch (type) {
+          case INT8:
+          case INT16:
+          case INT32:
+          case INT64:
+            Number record = (Number) partitionKey;
```
aviyehuda: What is the reason you convert it to a number? if all you do is to do record.toString(), than there is no need to convert it. Unless you also want to validate its value.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldAndTimeBasedPartitioner.java
```diff
@@ -0,0 +1,123 @@
...
+          Number record = (Number) fieldValue;
+          builder.append(record.toString());
+        } else if (fieldValue instanceof String) {
+          builder.append((String) fieldValue);
+        } else if (fieldValue instanceof Date) {
+          boolean booleanRecord = (boolean) fieldValue;
+          builder.append(Boolean.toString(booleanRecord));
+        } else {
+          log.error(
+                  "Unsupported type '{}' for user-defined timestamp field.",
```
jeremydonahue: > timestamp field
This should probably be _"field value"_.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldAndTimeBasedPartitioner.java
```diff
@@ -0,0 +1,123 @@
...
+                  fieldValue.getClass()
+          );
+          throw new PartitionException(
+                  "Error extracting timestamp from record field: " + fieldName
```
jeremydonahue: > Error extracting timestamp from...
This should probably be "Error extracting _value_ from..."

core/src/main/java/io/confluent/connect/storage/util/DataUtils.java
```diff
@@ -33,6 +33,7 @@ public static Object getField(Object structOrMap, String fieldName) {
       if (field == null) {
         throw new DataException(String.format("Unable to find nested field '%s'", fieldName));
       }
+      return field;
```
cricket007: This was fixed in #63

partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldAndTimeBasedPartitioner.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  @Override
+  public String encodePartition(SinkRecord sinkRecord) {
+    String timePartition = super.encodePartition(sinkRecord);
+
+    Object value = sinkRecord.value();
+    log.info(value.getClass().toString());
+
+    if (value instanceof Struct) {
```
atuljha-optimizely: Would `FieldPartitioner` be a better place for the field extraction logic, then we could simply return `fieldPartitioner.encodePartition(sinkRecord) + delim + super.encodePartition(sinkRecord)`

PR #88, opened 2018-10-30
CC-2617: Accept timestamp for wallclock instances to assure consistency

cyrusv:
Extending Partitioner and Extractor to accept a timestamp gives flexibility
for the calling function to inject desired behavior in some cases.

This has been particularly useful in getting consistent timestamps
in a wallclock based partitioners to return the same encodedPartition
for the same batch of records, preventing small files being created
across partition boundaries

### DIFF
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/Partitioner.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/Partitioner.java
@@ -32,9 +32,13 @@ public interface Partitioner<T> {
 
   /**
    * Returns string representing the output path for a sinkRecord to be encoded and stored.
+   *
    * @param sinkRecord The record to be stored by the Sink Connector
    * @return The path/filename the SinkRecord will be stored into after it is encoded
+   *
+   * @deprecated use {@link #encodePartition(SinkRecord, long)} instead
    */
+  @Deprecated
   String encodePartition(SinkRecord sinkRecord);
 
   /**
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/Partitioner.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/Partitioner.java
@@ -35,10 +35,7 @@ public interface Partitioner<T> {
    *
    * @param sinkRecord The record to be stored by the Sink Connector
    * @return The path/filename the SinkRecord will be stored into after it is encoded
-   *
-   * @deprecated use {@link #encodePartition(SinkRecord, long)} instead
    */
-  @Deprecated
   String encodePartition(SinkRecord sinkRecord);
 
   /**
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/Partitioner.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/Partitioner.java
@@ -32,6 +32,10 @@ public interface Partitioner<T> {
 
   String encodePartition(SinkRecord sinkRecord);
 
+  default String encodePartition(SinkRecord sinkRecord, long nowInMillis) {
+    return encodePartition(sinkRecord);
+  }
+
   String generatePartitionedPath(String topic, String encodedPartition);
 
   List<T> partitionFields();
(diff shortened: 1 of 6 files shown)


partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -152,9 +152,19 @@ public void configure(Map<String, Object> config) {
...
+    return encodedPartitionForTimestamp(sinkRecord, timestamp);
+  }
+
   @Override
   public String encodePartition(SinkRecord sinkRecord) {
     Long timestamp = timestampExtractor.extract(sinkRecord);
+    return encodedPartitionForTimestamp(sinkRecord, timestamp);
+  }
+
+  private String encodedPartitionForTimestamp(SinkRecord sinkRecord, Long timestamp) {
```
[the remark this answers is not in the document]
cyrusv: Since it is shared by both signatures of `encodePartition`

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -219,6 +229,26 @@ public TimestampExtractor newTimestampExtractor(String extractorClassName) {
     @Override
     public void configure(Map<String, Object> config) {}
 
+    /**
+     * If caller specifies nowInMillis, caller can better control and predict behavior of wallclock
+     * extractor. Can be useful to have records of the same batch extract to the same time for
+     * consistency.
```
rhauch: Suggest:
```
     * Returns the current timestamp supplied by the caller, which is assumed to be the processing time.
```

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -219,6 +229,26 @@ public TimestampExtractor newTimestampExtractor(String extractorClassName) {
...
+     * @param nowInMillis Time in ms specified by caller, useful for getting consistent wallclocks
+     * @return The wallclock specified by the input parameter in milliseconds
+     */
+    @Override
+    public Long extract(ConnectRecord<?> record, long nowInMillis) {
+      return nowInMillis;
+    }
+
+    /**
+     * Returns system time, aka Wallclock
```
rhauch:
```suggestion
     * Returns the current time from {@link Time#System Time.SYSTEM}.
```

partitioner/src/main/java/io/confluent/connect/storage/partitioner/Partitioner.java
```diff
@@ -30,8 +30,25 @@
 public interface Partitioner<T> {
   void configure(Map<String, Object> config);
 
+  /**
+   * Returns string representing the output path for a sinkRecord to be encoded and stored.
+   * @param sinkRecord The record to be stored by the Sink Connector
+   * @return The path/filename the SinkRecord will be stored into after it is encoded
```
rhauch: Add:
```
     * @deprecated use {@link #encodePartition(SinkRecord, long)} instead
```
and then add the annotation:
```
@Deprecated
```

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimestampExtractor.java
```diff
@@ -24,4 +24,8 @@
   void configure(Map<String, Object> config);
 
   Long extract(ConnectRecord<?> record);
```
rhauch: Please add JavaDoc, including:
```
     * @deprecated in favor of {@link #extract(ConnectRecord, long)}
```
and then add the annotation:
```
@Deprecated
```

partitioner/src/main/java/io/confluent/connect/storage/partitioner/Partitioner.java
cyrusv: I'd push back on this -- There are legitimate cases for not passing in the "now" time (e.g. a field-pased partitioner). Deprecating these creates a ton of warnings/failures because there are indeed places where it should be fine to call the old behavior (incl in the default function of the interface).

PR #89, opened 2018-11-06
Ability to use epoch and to also specify the format of a RecordField when using TimeBasedPartitioner

bwilliams42:
Hi. I added some functionality to use epoch seconds and to also specify the format of a `RecordField` when using `TimeBasedPartitioner`. For this I added two configuration attributes. 

`is.epoch` can be true but defaults to false. 
`timestamp.field.format` which expects a java date format string. 

I think it is useful and would like to share it. Thanks.

-BW

### DIFF
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
@@ -32,6 +32,7 @@ import io.confluent.connect.storage.common.ComposableConfig;
 
 public class PartitionerConfig extends AbstractConfig implements ComposableConfig {
 
+  public static final String IS_EPOCH = "is.epoch";
   // Partitioner group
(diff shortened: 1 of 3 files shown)


bwilliams42:
@kkonstantine happy Wednesday.I'm pinging you as I'm not sure what the protocol is. So if this is incorrect i apologize and ask that you let me know what the preferred process is. Thanks.

cricket007:
AFAIK, this feature set is already possible by putting a [`TimestampConverter$Value`](https://docs.confluent.io/current/connect/transforms/timestampconverter.html#timestampconverter) ahead of the Sink connector

piyushkhetan:
> AFAIK, this feature set is already possible by putting a [`TimestampConverter$Value`](https://docs.confluent.io/current/connect/transforms/timestampconverter.html#timestampconverter) ahead of the Sink connector

@cricket007 Thats true, but timestampConverter changes the value of the field. There may be cases where we don't want to update the field value and keep it in same format.

PR #102, opened 2019-04-30
CC-4433a: Remove Hive from packaging

kkonstantine:
* This PR removes hive dependencies from the `Format` interface and will allow connectors that depend on storage-common to selectively import Hive and Hadoop classes for integration with Hive only if they need to. 

* kafka-connect-storage-hive is removed from Confluent platform packages


package-kafka-connect-storage-common/pom.xml
```diff
@@ -47,11 +47,6 @@
             <artifactId>kafka-connect-storage-partitioner</artifactId>
             <version>${project.version}</version>
         </dependency>
-        <dependency>
-            <groupId>io.confluent</groupId>
-            <artifactId>kafka-connect-storage-hive</artifactId>
-            <version>${project.version}</version>
-        </dependency>
```
rhauch: Am I understanding it that we're removing this as a dependency, but we're keeping the `kafka-connect-storage-hive` module and code to only be including it in other modules (the HDFS connector) specifically via a dependency?

hive/src/main/java/io/confluent/connect/storage/hive/HiveConfig.java
```diff
@@ -155,26 +144,6 @@
...
-          group,
-          ++orderInGroup,
-          Width.SHORT,
-          SCHEMA_COMPATIBILITY_DISPLAY,
-          schemaCompatibilityRecommender
-      );
-    }
   }
 
   public static class SchemaCompatibilityRecommender extends BooleanParentRecommender {
```
rhauch: Is the intention to keep this class in `kafka-connect-storage-hive` even though it is also in `kafka-connect-storage-common` and the former depends on the latter, for backward compatibility reasons in case non-Confluent modules depend/use it?

package-kafka-connect-storage-common/pom.xml
rhauch: I think from https://github.com/confluentinc/kafka-connect-hdfs/pull/429 the answer to this is "yes".

kkonstantine: Indeed. This is removed from the packaging of Confluent platform, but will still be published in our public maven repo (nexus) for any other module, like the hdfs connector, to import it as a dependency

hive/src/main/java/io/confluent/connect/storage/hive/HiveConfig.java
kkonstantine: No, I just missed removing it. I feel that if we remove `HiveConfig.SCHEMA_COMPATIBILITY_CONFIG` we should remove this class too. The other option is to keep both with a note that we deprecate the ones here.

kkonstantine: WDYT?

rhauch: Sounds good.

PR #105, opened 2019-06-14
Upgrade Hadoop to 3.1.2 and Hive to 3.1.1

schroedermatt:


schroedermatt:
Moving a different direction with this. The deps will be updated in the HDFS3 connector project itself.

kkonstantine:
Indeed. This is not the right repo for this upgrade at this point @schroedermatt 
A connector can pin the version of the hadoop libraries directly.

PR #110, opened 2019-08-15
CC-5963: Improved ability to handle schema evolution with different names

rhauch:
Added additional checks to `StorageSchemaCompatibility.shouldChangeSchema(…)` so that this method considers not only the versions but also considers a schema as having changed when the name is changed, the types are different (but not projectable), or the parameters are changed. This makes this logic more compatible with the `project(…)` methods on the same enumeration used to project a record into the schema used in a file with lots of records.

Thus, connectors that use this will effectively rotate whenever such schema evolution occurs and projection (using Connect’s SchemaProjector) would not work successfully. Most users don’t have this kind of schema evolution, and thus would see no change in behavior. And users that do have this kind of schema evolution likely do so infrequently.

This change is still compatible with Schema Registry and Avro’s schema compatiblity behavior. In particular, Avro allows schemas to have different names and namespaces as long as the field requirements are still satisfied. So when a user is using SR and backward, forward, or full compatbility, SR and the Avro serializer will not prevent writing a record with an evolved schema. With this change, the records on a topic consumed by a connector using this class will consider the evolved schema as “changed”, and the connector will rotate its file by closing it, creating a new one, and writing the record with the new schema to the new file.

Quite a few unit tests were added to verify the StorageSchemaCompatibility behaviors, including the new behavior.

This PR backports this change to 4.0.x, since it is really a bugfix rather than an enhancement. Users that don't evolve their schemas to have different names or parameters will see no changes in behavior, as the old checks using version will still be performed. However, users that do evolve their schemas to have different names/namespaces will likely do so infrequently and thus will see the extra rotations infrequently.

### DIFF
--- a/core/src/main/java/io/confluent/connect/storage/schema/StorageSchemaCompatibility.java
+++ b/core/src/main/java/io/confluent/connect/storage/schema/StorageSchemaCompatibility.java
@@ -24,9 +24,13 @@ import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.source.SourceRecord;
 
 import java.util.AbstractMap;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.Set;
 
 public enum StorageSchemaCompatibility implements SchemaCompatibility {
   NONE {
@@ -48,6 +52,17 @@ public enum StorageSchemaCompatibility implements SchemaCompatibility {
       return record;
     }
 
+    /**
+     * Check whether the two schemas are incompatible such that they would prevent successfully
+     * {@link #project projecting} a key or value with the original schema into the current schema.
+     *
+     * <p>This method currently considers schemas to be compatible if and only they are
+     * {@link Schema#equals(Object) equal}. Otherwise, the schemas are deemed incompatible.
+     *
+     * @param originalSchema the original (new) schema; may not be null
+     * @param currentSchema  the current schema; may not be null
+     * @return true if the schemas are equals, or false if they are not equal
+     */
     @Override
     protected boolean check(
         Schema originalSchema,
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/confluent/connect/storage/schema/StorageSchemaCompatibility.java
```diff
@@ -58,8 +73,22 @@ protected boolean check(
...
+     * Check whether the two schemas have incompatible versions that prevent successfully
+     * {@link #project projecting} a key or value with the original schema into the current schema.
+     *
+     * <p>This method will consider schemas compatible for projection if the original schema's
+     * {@link Schema#version() version} is less than or equal to the current schema's version.
+     * IOW, if the original schema has an <em>older</em> version than the current schema, the
+     * schemas will be incompatible.
+     *
+     * @param originalSchema the original (new) schema; may not be null
+     * @param currentSchema  the current schema; may not b enull
```
mageshn: nit : may not be null

core/src/main/java/io/confluent/connect/storage/schema/StorageSchemaCompatibility.java
```diff
@@ -69,25 +98,54 @@ protected boolean check(
   FULL;
 
   private static final Map<String, StorageSchemaCompatibility> REVERSE = new HashMap<>();
+  private static final Set<SimpleImmutableEntry<Schema.Type, Schema.Type>> PROMOTABLES;
```
mageshn: Really wish we could use this from `SchemaProjector` rather than duplicating the code

core/src/main/java/io/confluent/connect/storage/schema/StorageSchemaCompatibility.java
```diff
@@ -105,13 +163,156 @@ protected boolean validateAndCheck(
...
+   * Check whether the two schemas have incompatible versions that prevent successfully
+   * {@link #project projecting} a key or value with the original schema into the current schema.
+   *
+   * <p>This method will consider schemas compatible for projection if the original schema's
+   * {@link Schema#version() version} is less than or equal to the current schema's version.
+   * IOW, if the original schema has a <em>newer</em> version than the current schema, the schemas
+   * will be incompatible.
+   *
+   * @param originalSchema the original (new) schema; may not be null
+   * @param currentSchema  the current schema; may not b enull
```
mageshn: nit: may not be null

PR #101, opened 2019-04-29
CC-147:make HiveSchemaConverter handle logical types

liukrimhrim:

PR #114, opened 2019-10-03
CC-147: logical types support for HiveSchemaConverter

liukrimhrim:
This is PR basically does what this old zombie PR does https://github.com/confluentinc/kafka-connect-storage-common/pull/101 . I made this new PR because the old one was based on the wrong branch and was targeting the wrong branch. 

Once merged, `HiveSchemaConverter` would be able to support logical types: decimal, timestamp, time and date.

### DIFF
--- a/hive/src/main/java/io/confluent/connect/storage/hive/HiveConfig.java
+++ b/hive/src/main/java/io/confluent/connect/storage/hive/HiveConfig.java
@@ -69,6 +69,16 @@ public class HiveConfig extends AbstractConfig implements ComposableConfig {
   public static final String SCHEMA_COMPATIBILITY_DEFAULT = "NONE";
   public static final String SCHEMA_COMPATIBILITY_DISPLAY = "Schema Compatibility";
 
+  //Logical type handling group
+  public static final String HIVE_LOGICAL_TYPES_ENABLED_CONFIG = "hive.logical.types.enabled";
+  public static final String HIVE_LOGICAL_TYPES_ENABLED_DOC = "Whether enable hive logical types:"
+      + "DECIMAL, DATE, INTERVAL, TIMESTAMP."
+      + "Please be noted that some of the logical types are not available "
+      + "in earlier versions of Hive. Better to disable hive logical type"
+      + "if using these earlier versions";
+  public static final boolean HIVE_LOGICAL_TYPES_ENABLED_DEFAULT = false;
+  public static final String HIVE_LOGICAL_TYPES_ENABLED_DISPLAY = "Enable Hive Logical Types";
+
   // CHECKSTYLE:OFF
   public static final ConfigDef.Recommender hiveIntegrationDependentsRecommender =
       new BooleanParentRecommender(HIVE_INTEGRATION_CONFIG);
@@ -173,6 +183,25 @@ public class HiveConfig extends AbstractConfig implements ComposableConfig {
           Width.SHORT,
           SCHEMA_COMPATIBILITY_DISPLAY,
           schemaCompatibilityRecommender
+      );
+    }
+
+    {
+      // Define Logical type handling configuration group
+      final String group = "Logical type handling";
+      int orderInGroup = 0;
+
+      // Define Logical type handling configuration group
+      CONFIG_DEF.define(
+          HIVE_LOGICAL_TYPES_ENABLED_CONFIG,
+          Type.BOOLEAN,
+          HIVE_LOGICAL_TYPES_ENABLED_DEFAULT,
+          Importance.MEDIUM,
+          HIVE_LOGICAL_TYPES_ENABLED_DOC,
+          group,
+          ++orderInGroup,
+          Width.SHORT,
+          HIVE_LOGICAL_TYPES_ENABLED_DISPLAY
       );
     }
   }
--- a/hive/src/main/java/io/confluent/connect/storage/hive/HiveConfig.java
+++ b/hive/src/main/java/io/confluent/connect/storage/hive/HiveConfig.java
@@ -71,12 +71,15 @@ public class HiveConfig extends AbstractConfig implements ComposableConfig {
 
   //Logical type handling group
   public static final String HIVE_LOGICAL_TYPES_ENABLED_CONFIG = "hive.logical.types.enabled";
-  public static final String HIVE_LOGICAL_TYPES_ENABLED_DOC = "Whether enable hive logical types:"
-      + "DECIMAL, DATE, INTERVAL, TIMESTAMP."
-      + "Please be noted that some of the logical types are not available "
-      + "in earlier versions of Hive. Better to disable hive logical type"
-      + "if using these earlier versions";
-  public static final boolean HIVE_LOGICAL_TYPES_ENABLED_DEFAULT = false;
+  public static final String HIVE_LOGICAL_TYPES_ENABLED_DOC = "Whether to enable logical "
+      + "type conversion: \n"
+      + "Connect Decimal -> Hive DECIMAL \n "
+      + "Connect Date -> Hive DATE \n "
+      + "Connect Timestamp -> Hive TIMESTAMP \n"
+      + "Connect Time is not supported with current implementation of Hive Serde."
+      + "Beware that the 3 logical types above are not compatible with "
+      + "some older versions of Hive.";
+  public static final boolean HIVE_LOGICAL_TYPES_ENABLED_DEFAULT = true;
   public static final String HIVE_LOGICAL_TYPES_ENABLED_DISPLAY = "Enable Hive Logical Types";
 
   // CHECKSTYLE:OFF
--- a/hive/src/main/java/io/confluent/connect/storage/hive/HiveConfig.java
+++ b/hive/src/main/java/io/confluent/connect/storage/hive/HiveConfig.java
@@ -69,19 +69,6 @@ public class HiveConfig extends AbstractConfig implements ComposableConfig {
   public static final String SCHEMA_COMPATIBILITY_DEFAULT = "NONE";
   public static final String SCHEMA_COMPATIBILITY_DISPLAY = "Schema Compatibility";
 
-  //Logical type handling group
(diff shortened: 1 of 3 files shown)


liukrimhrim:
@kkonstantine @wicknicks Would you mind having a look at this PR? Thank you!
Much appreciated for comments from any team members!

hive/src/main/java/io/confluent/connect/storage/hive/HiveConfig.java
```diff
@@ -70,6 +70,16 @@
   public static final String SCHEMA_COMPATIBILITY_DEFAULT = "NONE";
   public static final String SCHEMA_COMPATIBILITY_DISPLAY = "Schema Compatibility";
 
+  //Logical type handling group
+  public static final String HIVE_LOGICAL_TYPES_ENABLED_CONFIG = "hive.logical.types.enabled";
+  public static final String HIVE_LOGICAL_TYPES_ENABLED_DOC = "Whether enable hive logical types:"
```
levzem:
```suggestion
  public static final String HIVE_LOGICAL_TYPES_ENABLED_DOC = "Whether to enable hive logical types:"
```

hive/src/main/java/io/confluent/connect/storage/hive/HiveConfig.java
```diff
@@ -70,6 +70,16 @@
...
+      + "DECIMAL, DATE, INTERVAL, TIMESTAMP."
+      + "Please be noted that some of the logical types are not available "
```
levzem:
```suggestion
      + "Please note that some of the logical types are not available "
```

hive/src/main/java/io/confluent/connect/storage/hive/HiveConfig.java
```diff
@@ -70,6 +70,16 @@
...
+      + "in earlier versions of Hive. Better to disable hive logical type"
```
levzem:
```suggestion
      + "in earlier versions of Hive."
```

hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
```diff
@@ -27,11 +30,22 @@
...
+import org.apache.kafka.connect.errors.ConnectException;
 
 public class HiveSchemaConverter {
 
   private static final Map<Type, TypeInfo> TYPE_TO_TYPEINFO;
 
+  // the name has to be consistent with io.confluent.connect.avro.AvroData, when Connect Decimal
+  // schema is created, this property name is used to set precision.
+  // We have to use the exact name to retrieve precision value.
+  protected static final String CONNECT_AVRO_DECIMAL_PRECISION_PROP = "connect.decimal.precision";
```
levzem: can you just retrieve this name from AvroData?

hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
```diff
@@ -93,4 +131,36 @@ public static TypeInfo convertMap(Schema schema) {
...
+
+    switch (schema.name()) {
+      case Decimal.LOGICAL_NAME:
+        String scale = schema.parameters().get(Decimal.SCALE_FIELD);
+        String precision = schema.parameters().get(CONNECT_AVRO_DECIMAL_PRECISION_PROP);
+        if (precision == null || Integer.valueOf(precision) <= DECIMAL_PRECISION_DEFAULT) {
+          precision = String.valueOf(DECIMAL_PRECISION_DEFAULT);
+        } else {
+          throw new ConnectException("Illegal precision: "
+              + "Because Hive allows at most 38 precision, the precision"
```
levzem:
```suggestion
              + "Because Hive allows at most DECIMAL_PRECISION_DEFAULT precision, the precision"
```

hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
kkonstantine: it's not visible. It's package private.

hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
```diff
@@ -27,11 +30,22 @@
...
 
   private static final Map<Type, TypeInfo> TYPE_TO_TYPEINFO;
 
+  // the name has to be consistent with io.confluent.connect.avro.AvroData, when Connect Decimal
+  // schema is created, this property name is used to set precision.
+  // We have to use the exact name to retrieve precision value.
+  protected static final String CONNECT_AVRO_DECIMAL_PRECISION_PROP = "connect.decimal.precision";
+
+  // this is the maximum digits Hive allows for DECIMAL type.
+  protected static final int DECIMAL_PRECISION_DEFAULT = 38;
```
kkonstantine: Let's call this `HIVE_DECIMAL_PRECISION_MAX`
to distinguish it from `AvroData#CONNECT_AVRO_DECIMAL_PRECISION_DEFAULT` or similar
and also note that this is not a default of a configurable option but rather the max of a configurable option

hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
kkonstantine: Please accept recommendation but with the correct name. Actually, even better, don't use the string name of the config, but use the config as variable. In that case you have the best of both options. You always have a value in the config, and that value is always the correct one.

hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
```diff
@@ -93,4 +131,36 @@ public static TypeInfo convertMap(Schema schema) {
...
+    switch (schema.name()) {
+      case Decimal.LOGICAL_NAME:
+        String scale = schema.parameters().get(Decimal.SCALE_FIELD);
+        String precision = schema.parameters().get(CONNECT_AVRO_DECIMAL_PRECISION_PROP);
+        if (precision == null || Integer.parseInt(precision) <= DECIMAL_PRECISION_DEFAULT) {
+          precision = String.valueOf(DECIMAL_PRECISION_DEFAULT);
+        } else {
+          throw new ConnectException("Illegal precision: "
+              + "Because Hive allows at most 38 precision, the precision"
+              + "has to be less than 39.");
```
kkonstantine: remember to substitute the variable here too

hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
```diff
@@ -93,4 +131,36 @@ public static TypeInfo convertMap(Schema schema) {
...
+    if (schema.name() == null) {
+      return convertPrimitive(schema);
+    }
+
+    switch (schema.name()) {
+      case Decimal.LOGICAL_NAME:
+        String scale = schema.parameters().get(Decimal.SCALE_FIELD);
+        String precision = schema.parameters().get(CONNECT_AVRO_DECIMAL_PRECISION_PROP);
+        if (precision == null || Integer.parseInt(precision) <= DECIMAL_PRECISION_DEFAULT) {
+          precision = String.valueOf(DECIMAL_PRECISION_DEFAULT);
```
kkonstantine: Are we risking data loss if `precision == null`?

liukrimhrim: No, we are not. Actually, I don't need to check all this booleans here, I can just make precision = 38 all the time. I've check Hive's serde code, it will do all the logic trying the best to fit decimal data into decimal schema. If the data is too long to fit in, hive will throw SerdeException.

liukrimhrim:
@kkonstantine , @levzem thanks for the review. After struggling with this ticket for a month, I finally have a clear thought of how to resolve it:
1. To support logical type, two things have to be achieved: one is that the corresponding Hive column schemas have to be of the correct hive type (might be Hive logical types); the other one is that hive serde implementation supports Ser/De data of logical types.
2. Hive is built on top of HDFS, and file format can be avro or parquet. 
3. If it is avro, I think we don't need to do anything for logical type support. We only need to set avro schema literal as a property of hive table (which we already did), hive table will convert connect logical types to hive logical types in table schema. And hive serde can correctly Ser/De avro files.
4. If it is parquet, we need to manually convert connect logical types to hive logical types in schema. And hive serde will do the rest of the work. 
5. Decimal is a special logical type because it has precision and scale, and there is a chance that some coming data might not fit the schema. Moreover, Hive puts a cap on precision (which is 38), but connect allows any precision. Hive serde will try the best to fit decimal data to schema, but if the decimal is just too long to fit in, it will throw serde exception.
6. Connect time type cannot be supported with current implementation of hive serde.

So, my fix would be: add an option to hive config to turn ON/OFF hive logical type support. If it is ON, and the file format is avro, I will add extra logic to prevent default logical type support (as I mentioned in 3.). If it is OFF, and the file format is parquet, I will use new logic added to `HiveSchemaConverter` to convert connect logical types to hive logical types. For decimal types, I will make the precision as large as hive allows, and hive serde will try the best to fit data to the schema. If the decimal is just too long to fit, hive will throw exception. So, no risk of silent data loss.

I have to issue two PRs. This is the first one which adds extra logic to `HiveSchemaConverter` to convert connect logical types to hive logical types. The PR will be on HDFS connector, which adds new logic to decide behavior under different logical type support ON/OFF + Avro/Parquet scenarios

kkonstantine:
Makes sense @liukrimhrim . Thanks for the detailed explanation. 
I think we are ok here, but I'd like to first review the follow up PRs that depend on this PR before we merge here (introducing a new config)

kkonstantine:
Oh, and one thing I forgot. Would it make sense to get some more unit tests here for the other logical types?

liukrimhrim:
> Oh, and one thing I forgot. Would it make sense to get some more unit tests here for the other logical types?

There are tests testing other logical types already

liukrimhrim:
Retargeted to 5.1.x because it depends on avro-converter 5.1.x and later.

PR #116, opened 2019-10-25
CC-6565: Add parquet.codec property to enable compression in storage sink connectors

kkonstantine:
This configuration property will allow storage sink connectors that support Parquet to compress their output files. 

The signature of the internal method newConfigDef is also changing here and the storage sink connectors that use it will have to adapt to this change.

### DIFF
--- a/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
+++ b/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
@@ -118,6 +118,12 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
   public static final String[] AVRO_SUPPORTED_CODECS = new String[]{"null", "deflate", "snappy",
       "bzip2"};
 
+  public static final String PARQUET_CODEC_CONFIG = "parquet.codec";
+  public static final String PARQUET_CODEC_DEFAULT = "snappy";
+  public static final String PARQUET_CODEC_DISPLAY = "Parquet Compression Codec";
+  public static final String PARQUET_CODEC_DOC = "The Parquet compression codec to be used for "
+      + "output files.";
+
   // Schema group
   public static final String SCHEMA_COMPATIBILITY_CONFIG = "schema.compatibility";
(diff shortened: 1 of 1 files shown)


kkonstantine:
@zzbennett thx! Good catch. 
Also changed the default to be `snappy` because that's what the default has been for hdfs connector.

rhauch:
Is `master` the correct branch for this PR? BTW, even if it is the correct branch, we're branching today, so we might need to retarget this to `5.4.x` instead.

kkonstantine:
Thanks @rhauch. Indeed, creation of 5.4.x is underway, so I'll wait till it's created and will retarget this PR.

rhauch:
@kkonstantine, it occurred to me that not every connector that uses this library may want to support Parquet. WDYT about having a separate static method add the parquet config to a supplied ConfigDef? That would allow connectors to upgrade to `kafka-connect-storage-common:5.4.0` and optionally enable Parquet.

kkonstantine:
Good point @rhauch 
I updated the PR to introduce an optional static method. Let me know what you think. 
I've adjusted the PR on S3 to show how this method can be used now: https://github.com/confluentinc/kafka-connect-storage-cloud/pull/282

kkonstantine:
If this now looks good, I'll rebase on top of `5.4.x` before I merge.

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
```diff
@@ -353,6 +390,10 @@ public String getAvroCodec() {
     return getString(AVRO_CODEC_CONFIG);
   }
 
+  public String parquetCodec() {
+    return getString(PARQUET_CODEC_CONFIG);
+  }
```
rhauch: This method will throw an exception if the config key was not defined. How about either documenting that or checking whether it is defined first?

kkonstantine:
@rhauch I removed `parquetCodec`. It was added to be symmetric to `avroCodec` and is not used a log. I left it because I thought anyone who'd use it would know why, but I agree, we are better without it, given that the config might not be present.

PR #123, opened 2020-02-03
[6.0] CC-3543: Update Hadoop and Hive versions

ncliang:
This commit bumps Hadoop version to 2.10.0 released 10/2019 and Hive version to 2.3.6 released 8/2019.

Muckrake system test run #3352 is passing with this change.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -174,7 +174,7 @@
                 <configuration>
                     <showWarnings>true</showWarnings>
                     <compilerArgs>
-                        <arg>-Xlint:all</arg>
+                        <arg>-Xlint:-processing</arg>
                         <arg>-Werror</arg>
                     </compilerArgs>
                 </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -67,8 +67,8 @@
         <commons-io.version>2.4</commons-io.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
         <confluent.version>6.0.0-SNAPSHOT</confluent.version>
-        <hadoop.version>2.7.3</hadoop.version>
-        <hive.version>1.2.2</hive.version>
+        <hadoop.version>2.10.0</hadoop.version>
+        <hive.version>2.3.6</hive.version>
         <joda.version>2.9.6</joda.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -67,8 +67,8 @@
         <commons-io.version>2.4</commons-io.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
         <confluent.version>5.5.0-SNAPSHOT</confluent.version>
-        <hadoop.version>2.7.3</hadoop.version>
-        <hive.version>1.2.2</hive.version>
+        <hadoop.version>2.10.0</hadoop.version>
+        <hive.version>2.3.6</hive.version>
```
wicknicks: this is a major version bump for hive. do we meet all API compatibility requirements with existing code?

ncliang: I am not sure I understand what you mean by all API compatibility requirements? Muckrake system tests pass, indicating at least what API we use in those tests still behave in the same way we expect. Downstream `kafka-connect-hdfs` project has tests using embedded miniDC that require some tweaks to please the miniDC test framework, but no production code changes. In fact, we already do use hive version 3 in the hdfs3 connector and v2.3.5 in the GCP Dataproc connector.

With such dependency version bumps, there will always be a risk of us depending on behavior that has changed in the new version. I feel that the risk here is justified as there are multiple CVEs identified in dependencies pulled in by older versions of hadoop/hive. CC-3543 and other JIRAs linked or mentioned there contain more details.

Let me know if you would like to see more test coverage to feel more confident about the version bump? I am separately making a change to test code in HDFS repository to pick up the version bump there.

ncliang:
@wicknicks I have rebased the changes onto master post release branch for 5.5.x. They should land and be released together with 6.0. LMK if you have other concerns around the version bump.

wicknicks:
@ncliang in the README docs for CP 5.5, we might want to indicate that v6.0 of the HDFS connector will use the latest libraries for hive (1.x -> 2.y), which breaks usability with very old hadoop clusters (1.x and even 2.x, IIRC). there is a section for deprecation notices such as this.

PR #125, opened 2020-03-01
Revert "[6.0] CC-3543: Update Hadoop and Hive versions (#123)"

C0urante:
This reverts commit 286b9b80e0fbac7ef98e6c01eba26e20c853491c.

This change has been causing downstream build failures with the HDFS connector, which are in turn blocking some of our other downstream builds. We should revert these changes for now and only merge them back after verifying that they work with the HDFS connector.

### DIFF
--- a/hive/pom.xml
+++ b/hive/pom.xml
@@ -32,7 +32,6 @@
     <properties>
         <commons.lang3.version>3.1</commons.lang3.version>
         <kryo.version>2.22</kryo.version>
-        <xml-apis.version>1.4.01</xml-apis.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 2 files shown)


ncliang:
Do we need to revert? I have a PR in HDFS repo that fixes the build break.
I was going to land then together yesterday but got distracted by an
escalation. I can land that change tonight to fix the build break.

C0urante:
@ncliang do you mean https://github.com/confluentinc/kafka-connect-hdfs/pull/486? Because that PR is failing, too.

ncliang:
That was passing, but was failing because I was testing whether some test
changes were required. It will pass once I put back the test changes I
removed.

C0urante:
@ncliang We should fix the build for the HDFS connector as quickly as possible to avoid blocking downstream builds that rely on it; if your fix works then it's totally fine (and probably preferable) to merge it now instead of this PR.

C0urante:
Closing as the merge and successful build of https://github.com/confluentinc/kafka-connect-hdfs/pull/486 has made this unnecessary.

Issue #121, opened 2020-01-26
add a property to tell timestamp extractor whether timestamp value is in seconds or milliseconds


bagipriyank:
timestamp extractor assumes that timestamp value is in seconds which is causing an issue  mentioned in https://github.com/confluentinc/kafka-connect-storage-common/issues/120#issuecomment-578459085 

adding a property to tell timestamp extractor that timestamp is in seconds will help resolve this issue.

nojetlag:
in general this should be much more flexible especially with financial data often timestamps are in microseconds.

PR #122, opened 2020-01-27
added a timestamp unit to partitioner config

bagipriyank:

PR #127, opened 2020-03-05
Improvements to storage partitioning scheme to be flexible and configurable at runtime.

hariprasad-k:
Improvements:

- Ability to exclude topic name in partiton by configuration
- Ability to specify scaling factor for timestamp column used in partitioning by configuration


hariprasad-k:
Fixed #127

hariprasad-k:
The patch also fixes #121 too

cricket007:
Question: What happens when you have `topics` as a list or `topics.regex` capture more than one topic?

Then all files in the directory will be of multiple topics?

partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
```diff
@@ -101,6 +101,20 @@
   public static final String TIMESTAMP_FIELD_NAME_DEFAULT = "timestamp";
   public static final String TIMESTAMP_FIELD_NAME_DISPLAY = "Record Field for Timestamp Extractor";
 
+  public static final String TIMESTAMP_SCALING_FACTOR_CONFIG = "timestamp.scaling.factor";
```
cricket007: I suggest you break this out as a separate PR

cricket007:
> The patch also fixes #121 too

So does #122

hariprasad-k:
> Question: What happens when you have `topics` as a list or `topics.regex` capture more than one topic?
> 
> Then all files in the directory will be of multiple topics?

Yes, all files of multiple topics will be in the same directory in this case.

But, this behavior will only happen if you use `topics` as a list or `topics.regex` in combination with `path.include.topicname` set as `false`. Further, as `path.include.topicname` is by default `true`, it will preserve the existing behavior by default.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/PartitionerConfig.java
hariprasad-k: @OneCricketeer Sure, I will break into separate PR.

hariprasad-k:
@OneCricketeer This PR does not have any changes related to Timestamp scaling factor anymore. 

Can you please review this changes ?

OneCricketeer:
I'm not a Confluent employee, so my reviews don't do anything

dosvath:
Thanks @hariprasad-k for the contribution! This change will add flexibility to our connectors that use storage common to optionally exclude the topic name (seems to be a popular request https://github.com/confluentinc/kafka-connect-hdfs/issues/544). I'd like to get second view from @levzem before proceeding.

hariprasad-k:
@dosvath @levzem Any thoughts, or plans to review this contribution ?

PR #133, opened 2020-05-16
Added complex key/header generation to SchemaSourceTask

dosvath:
Adding `complex.keys.headers` configuration to be able to generate complex record `keys` and `headers` for use in tests.

### DIFF
--- a/core/src/main/java/io/confluent/connect/storage/tools/SchemaSourceTask.java
+++ b/core/src/main/java/io/confluent/connect/storage/tools/SchemaSourceTask.java
@@ -19,6 +19,8 @@ import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.header.ConnectHeaders;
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/confluent/connect/storage/tools/SchemaSourceTask.java
```diff
@@ -86,6 +91,54 @@
       .field("seqno", Schema.INT64_SCHEMA)
       .build();
 
+  private static Schema simpleKeySchema = Schema.STRING_SCHEMA;
+  private static Schema complexKeySchema = SchemaBuilder.struct().name("complexkeyschema")
```
levzem, line 95:
```suggestion
  private static Schema complexKeySchema = SchemaBuilder.struct().name("complex-key-schema")
```

dosvath: This is a tricky one, I had to remove the dashes because tests were failing; it looks like field names in avro schemas don't support special characters. 
` ERROR Could not parse Avro schema (io.confluent.kafka.schemaregistry.avro.AvroSchemaProvider:43)
org.apache.avro.SchemaParseException: Illegal character in: complex-key-schema` https://avro.apache.org/docs/1.7.7/spec.html#Names

PR #138, opened 2020-07-10
CC-12005: Exclude vulnerable dependency jackson mapper asl

Prathibha-m:
UPDATE: The fix was added (dependency was excluded) in gcs,azure blob and hdfs connectors. We should make a call whether this fix needs to be in storage-common
## Problem
Resolving connector CVE for jackson-mapper-asl

## Solution
straight forward exclusion of the transitive dependency, since it isn't used by the connector.

- [x] yes
- [ ] no

##### If yes, where?
HDFS sink and GCS sink

## Test Strategy
Branch to be tested against Muckrake.

Understand testing for all connectors that use `kafka-connect-storage-common` for these fixes

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/hive/pom.xml
+++ b/hive/pom.xml
@@ -55,6 +55,12 @@
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-client</artifactId>
             <version>${hadoop.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.codehaus.jackson</groupId>
+                    <artifactId>jackson-mapper-asl</artifactId>
+                </exclusion>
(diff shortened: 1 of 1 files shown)


dosvath:
@kkonstantine do we want to target master here?

kkonstantine (commented):
@Prathibha-m we'll need to resolve conflicts and also run the system tests with the suggested changes to make sure things don't break after the exclusions.

kkonstantine (approved):
LGTM but please target 10.0.x

Issue #137, opened 2020-07-02
Incompatibility between parquet 1.11.0 and apache/arrow (snowflake db)


BDeus:
We use the parquet format to provide data to a Snowflake DB (with s3 connector). 
Since the bump of parquet to 1.11.0, we have incompatibility issue (with map type and avro parquet writer).

It seems a patch will be available soon. Could you upgrade to the lastest version of parquet as soon as it becomes available ?
Thanks.
[StackOverflow Link](https://stackoverflow.com/questions/62504757/issue-with-loading-parquet-data-into-snowflake-cloud-database-when-written-with)
[JIRA Link](https://issues.apache.org/jira/browse/PARQUET-1879?jql=project%20%3D%20PARQUET%20AND%20resolution%20%3D%20Unresolved%20AND%20text%20~%20%22snowflake%22%20ORDER%20BY%20priority%20ASC%2C%20updated%20DESC)
[Parquet PR Link](https://issues.apache.org/jira/browse/PARQUET-1879?jql=project%20%3D%20PARQUET%20AND%20resolution%20%3D%20Unresolved%20AND%20text%20~%20%22snowflake%22%20ORDER%20BY%20priority%20ASC%2C%20updated%20DESC)

BDeus:
The parquet version 1.11.1 is available. 
Can someone upgrade the version in the pom.xml to patch the bug present on avro map ?
Thanks

BDeus:
Confirm that with 1.11.1 it's ok now

PR #148, opened 2020-08-31
Upgrade parquet.version to 1.11.1

BDeus:
Correct the issue #137


levzem:
@BDeus can you please backport the change to when `parquet 1.11.0` was first introduced? otherwise LGTM!

PR #150, opened 2020-09-10
CC-11532: TimeBasedPartitioner to extract timestamp from a string literal as unixtime or ISO-8601-formatted date.

avocader:
## Problem

Currently, if S3 sink connector is configured as follows:

```
"connector.class":"io.confluent.connect.s3.S3SinkConnector",
"timestamp.extractor":"RecordField",
"timestamp.field":"ts.field",
"partitioner.class":"io.confluent.connect.storage.partitioner.TimeBasedPartitioner",
...
```

It can handle records, where the timestamp field has a string literal with ISO-8601-formatted date:
```
{"ts.field":"2000-08-20T4:05:30.799044","value":"xyz"}
```
But fails to handle a record, where the timestamp field has a string literal with Unix time:
```
{"ts.field":"1597904371234","value":"xyz"}
```

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
@@ -48,7 +48,8 @@ public class TimeBasedPartitioner<T> extends DefaultPartitioner<T> {
 
   private static final String SCHEMA_GENERATOR_CLASS =
       "io.confluent.connect.storage.hive.schema.TimeBasedSchemaGenerator";
-  static Pattern NUMERIC_TIMESTAMP_PATTERN = Pattern.compile("^-?[0-9]{1,19}$");
+  private static final Pattern NUMERIC_TIMESTAMP_PATTERN = Pattern.compile("^-?[0-9]{1,19}$");
+
 
   // Duration of a partition in milliseconds.
   private long partitionDurationMs;
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
@@ -15,7 +15,6 @@
 
 package io.confluent.connect.storage.partitioner;
 
-import io.confluent.connect.storage.util.DataUtils;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.connect.connector.ConnectRecord;
@@ -41,6 +40,7 @@ import io.confluent.connect.storage.common.SchemaGenerator;
 import io.confluent.connect.storage.common.StorageCommonConfig;
 import io.confluent.connect.storage.common.util.StringUtils;
 import io.confluent.connect.storage.errors.PartitionException;
+import io.confluent.connect.storage.util.DataUtils;
 
 public class TimeBasedPartitioner<T> extends DefaultPartitioner<T> {
   private static final Logger log = LoggerFactory.getLogger(TimeBasedPartitioner.class);
@@ -290,7 +290,7 @@ public class TimeBasedPartitioner<T> extends DefaultPartitioner<T> {
           case INT64:
             return ((Number) timestampValue).longValue();
           case STRING:
-            return dateTime.parseMillis((String) timestampValue);
+            return extractTimestampFromString(timestampValue);
           default:
             log.error(
                 "Unsupported type '{}' for user-defined timestamp field.",
@@ -306,7 +306,7 @@ public class TimeBasedPartitioner<T> extends DefaultPartitioner<T> {
         if (timestampValue instanceof Number) {
           return ((Number) timestampValue).longValue();
         } else if (timestampValue instanceof String) {
-          return dateTime.parseMillis((String) timestampValue);
+          return extractTimestampFromString(timestampValue);
         } else if (timestampValue instanceof Date) {
           return ((Date) timestampValue).getTime();
         } else {
@@ -322,6 +322,16 @@ public class TimeBasedPartitioner<T> extends DefaultPartitioner<T> {
         log.error("Value is not of Struct or Map type.");
         throw new PartitionException("Error encoding partition.");
       }
+    }
+
+    private Long extractTimestampFromString(Object timestampValue) {
+      String timestampValueStr = (String) timestampValue;
+      try {
+        return Long.valueOf(timestampValueStr);
+      } catch (NumberFormatException e) {
+        // expected, ignore
+      }
+      return dateTime.parseMillis(timestampValueStr);
     }
   }
 }
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
@@ -324,7 +324,7 @@ public class TimeBasedPartitioner<T> extends DefaultPartitioner<T> {
       }
     }
 
-    private Long extractTimestampFromString(Object timestampValue) {
+    private Long extractTimestampFromString(String timestampValue) {
       String timestampValueStr = (String) timestampValue;
       try {
         return Long.valueOf(timestampValueStr);
--- a/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
+++ b/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
@@ -48,7 +48,7 @@ public class TimeBasedPartitioner<T> extends DefaultPartitioner<T> {
 
   private static final String SCHEMA_GENERATOR_CLASS =
       "io.confluent.connect.storage.hive.schema.TimeBasedSchemaGenerator";
-  static Pattern NUMERIC_TIMESTAMP_PATTERN = Pattern.compile("^[0-9]{13}$");
+  static Pattern NUMERIC_TIMESTAMP_PATTERN = Pattern.compile("^-?[0-9]{1,19}$");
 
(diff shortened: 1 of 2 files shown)


partitioner/src/test/java/io/confluent/connect/storage/partitioner/TimeBasedPartitionerTest.java
```diff
@@ -583,6 +580,37 @@ public void testRecordFieldTimeStringExtractor() {
     validateEncodedPartition(encodedPartition);
   }
 
+  @Test
+  public void testRecordFieldTimeStringExtractorRecord() {
```
levzem, line 584:
```suggestion
  public void testRecordFieldTimeAsMillisStringExtractorRecord() {
```

partitioner/src/test/java/io/confluent/connect/storage/partitioner/TimeBasedPartitionerTest.java
```diff
@@ -583,6 +580,37 @@ public void testRecordFieldTimeStringExtractor() {
...
+    Schema schema = SchemaBuilder.struct().name("record")
+        .field(timeFieldName, Schema.STRING_SCHEMA);
+    Struct s = new Struct(schema).put(timeFieldName, timeStr);
+    SinkRecord sinkRecord = createValuedSinkRecord(schema, s, millis);
+    Long extractedTimestamp = partitioner.getTimestampExtractor().extract(sinkRecord);
+    assertThat(extractedTimestamp, is(millis));
+  }
+
+  @Test
+  public void testRecordFieldTimeStringExtractorMap() {
```
levzem, line 599:
```suggestion
  public void testRecordFieldTimeAsMillisStringExtractorMap() {
```

partitioner/src/test/java/io/confluent/connect/storage/partitioner/TimeBasedPartitionerTest.java
levzem: would be nice to rename the existing StringExtractor tests to indicate it is for ISO dates as well

levzem (approved):
LGTM with nits

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -323,5 +323,15 @@ public Long extract(ConnectRecord<?> record) {
         throw new PartitionException("Error encoding partition.");
       }
     }
+
+    private Long extractTimestampFromString(Object timestampValue) {
+      String timestampValueStr = (String) timestampValue;
+      try {
+        return Long.valueOf(timestampValueStr);
+      } catch (NumberFormatException e) {
```
kkonstantine, line 331: My main (big) concern is now we pay a tax when converting timestamps that are not of type long. 
The branching based on exception handling will be way more expensive. 

The best I can think of and confirmed with a research online is to use a regex. We probably don't want to introduce a dep to commons-lang (or similar) just for that.

kkonstantine (commented):
Thanks @avocader for extending this

Yet, the current approach seems to introduce overhead to the current main use case (string timestamps that are not longs). No easy/perfect solution since there's no java library method to tell us whether a string is a long. Should we look at compiling a regex instead? (there will be some overhead but much less).

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
levzem: We can just easily avoid this issue by catching the exception from `dateTime.parseMillis(timestampValueStr)` as that is the more common use case and that way we don't "pay the tax" (most of the time).

avocader: I did some performance testing, to understand the performance tax, that we are introducing.

Base case(handling ISO ts only):
|Number of iterations|ISO ts extraction runtime, ms|
|---------------------|-----------------------------|
| 1000000                   | 1385                                        | 
| 10000000                 | 12172                                      | 
| 100000000               | 121982                                   | 

No regex:
|Number of iterations|ISO ts extraction runtime, ms|Numeric ts extraction runtime, ms|
|---------------------|-----------------------------|---------------------------------|
| 1000000                   | 4253                                        | 45                                                  |
| 10000000                 | 41792                                      | 305                                                |
| 100000000               | 437195                                   |  2855                                             |

Using a regex:
|Number of iterations|ISO ts extraction runtime, ms|Numeric ts extraction runtime, ms|
|---------------------|-----------------------------|---------------------------------|
| 1000000                   | 1962                                        | 273                                                |
| 10000000                 | 13295                                     | 1885                                               |
| 100000000               | 123757                                   | 15437                                             |

Looks like using a regex makes the numeric timestamp extraction slower, but keeps existing ISO timestamp extraction within a small delta.

partitioner/src/test/java/io/confluent/connect/storage/partitioner/TimeBasedPartitionerTest.java
```diff
@@ -583,6 +580,37 @@ public void testRecordFieldTimeStringExtractor() {
...
+
+  @Test
+  public void testRecordFieldTimeAsMillisStringExtractorMap() {
+    long millis = DATE_TIME.getMillis();
+    String timeStr = String.valueOf(millis);
+    String timeFieldName = "timestamp";
+    TimeBasedPartitioner<String> partitioner = configurePartitioner(
+        new TimeBasedPartitioner<>(), timeFieldName, null);
+    Schema keySchema = Schema.STRING_SCHEMA;
+    Schema mapSchema = SchemaBuilder.map(keySchema, Schema.STRING_SCHEMA);
```
levzem, line 606:
```suggestion
    Schema mapSchema = SchemaBuilder.map(Schema.STRING_SCHEMA, Schema.STRING_SCHEMA);
```

partitioner/src/test/java/io/confluent/connect/storage/partitioner/TimeBasedPartitionerTest.java
```diff
@@ -583,6 +580,37 @@ public void testRecordFieldTimeStringExtractor() {
...
+    Struct s = new Struct(schema).put(timeFieldName, timeStr);
+    SinkRecord sinkRecord = createValuedSinkRecord(schema, s, millis);
+    Long extractedTimestamp = partitioner.getTimestampExtractor().extract(sinkRecord);
+    assertThat(extractedTimestamp, is(millis));
+  }
+
+  @Test
+  public void testRecordFieldTimeAsMillisStringExtractorMap() {
+    long millis = DATE_TIME.getMillis();
+    String timeStr = String.valueOf(millis);
```
levzem, line 601:
```suggestion
    String timeStr = String.valueOf(DATE_TIME.getMillis());
```

partitioner/src/test/java/io/confluent/connect/storage/partitioner/TimeBasedPartitionerTest.java
```diff
@@ -583,6 +580,37 @@ public void testRecordFieldTimeStringExtractor() {
     validateEncodedPartition(encodedPartition);
   }
 
+  @Test
+  public void testRecordFieldTimeAsMillisStringExtractorRecord() {
+    long millis = DATE_TIME.getMillis();
+    String timeStr = String.valueOf(millis);
```
levzem, line 586: as below

levzem (approved):
few nits but LGTM, thanks @avocader

avocader:
@kkonstantine the current solution, that is in PR - it compiles a regex 
```
static Pattern NUMERIC_TIMESTAMP_PATTERN = Pattern.compile("^[0-9]{13}$");`
```
Is my understanding right, or you mean something else?

partitioner/src/test/java/io/confluent/connect/storage/partitioner/TimeBasedPartitionerTest.java
avocader: It's used below in two places, prefer to keep it as is.

kkonstantine (commented):
Thanks for the follow up @avocader 
Regex seems to make the most sense in order to have the same method covering both the long and the iso formats. The perf numbers seem reasonable. 

Couple of final suggestions to get the regex right.

partitioner/src/test/java/io/confluent/connect/storage/partitioner/TimeBasedPartitionerTest.java
```diff
@@ -583,6 +580,36 @@ public void testRecordFieldTimeStringExtractor() {
     validateEncodedPartition(encodedPartition);
   }
 
+  @Test
```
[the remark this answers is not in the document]
avocader: Sure, added more tests.

partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java
```diff
@@ -36,17 +35,20 @@
...
 import io.confluent.connect.storage.common.util.StringUtils;
 import io.confluent.connect.storage.errors.PartitionException;
+import io.confluent.connect.storage.util.DataUtils;
 
 public class TimeBasedPartitioner<T> extends DefaultPartitioner<T> {
   private static final Logger log = LoggerFactory.getLogger(TimeBasedPartitioner.class);
 
   private static final String SCHEMA_GENERATOR_CLASS =
       "io.confluent.connect.storage.hive.schema.TimeBasedSchemaGenerator";
+  static Pattern NUMERIC_TIMESTAMP_PATTERN = Pattern.compile("^[0-9]{13}$");
```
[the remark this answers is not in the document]
avocader: Updated the regex

kkonstantine (approved):
Thanks @avocader !
LGTM

PR #158, opened 2020-11-02
Reshade transitive dependency htrace-core4 with newer jackson to remove CVEs

kkonstantine:
## Problem
The HDFS connector (Hadoop v2) depends on a few transitive dependencies that are old and are shading a version of jackson that has CVEs associated with it. 

## Solution
We are reshading this dependency by removing the old jackson version and replacing it with a newer one that does not have CVEs. HDFS connector can import this dependency and get a clean htrace-core4 package. 

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/htrace-core4-shaded/pom.xml
+++ b/htrace-core4-shaded/pom.xml
@@ -1,0 +1,104 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+<!-- Licensed to the Apache Software Foundation (ASF) under one or more contributor
+license agreements. See the NOTICE file distributed with this work for additional
+information regarding copyright ownership. The ASF licenses this file to
+You under the Apache License, Version 2.0 (the "License"); you may not use
+this file except in compliance with the License. You may obtain a copy of
+the License at http://www.apache.org/licenses/LICENSE-2.0 Unless required
+by applicable law or agreed to in writing, software distributed under the
+License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS
+OF ANY KIND, either express or implied. See the License for the specific
+language governing permissions and limitations under the License. -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.confluent</groupId>
+        <artifactId>kafka-connect-storage-common-parent</artifactId>
+        <version>10.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>kafka-connect-storage-common-htrace-core4-shaded</artifactId>
+    <name>kafka-connect-storage-common-htrace-core4-shaded</name>
+    
+    <description>htrace-core4 shaded to replace jackson dependencies without CVEs</description>
+    <inceptionYear>2020</inceptionYear>
+
+    <properties>
+        <htrace.version>4.1.0-incubating</htrace.version>
+        <jackson.version>2.10.5</jackson.version>
+    </properties>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


htrace-core4-shaded/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+
+    <artifactId>kafka-connect-storage-common-htrace-core4-shaded</artifactId>
+    <name>kafka-connect-storage-common-htrace-core4-shaded</name>
+    
+    <description>htrace-core4 shaded to replace jackson dependencies without CVEs</description>
+    <inceptionYear>2020</inceptionYear>
+
+    <properties>
+        <htrace.version>4.1.0-incubating</htrace.version>
+        <jackson.version>2.10.5</jackson.version>
```
rhauch, line 30: What Jackson version is shaded in the original library? And how will we gain confidence that there are no compatibility problems by replacing that version with 2.10.5?

Should this use a variable defined in the parent POM? I guess maybe not, since it doesn't have to be the same Jackson version used by the connector since this is shaded and used only by the `htrace-core4` library.

htrace-core4-shaded/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+                                    <artifact>org.apache.htrace:htrace-core4</artifact>
+                                    <includes>
+                                        <include>**</include>
+                                    </includes>
+                                    <excludes>
+                                        <exclude>META-INF/MANIFEST.MF</exclude>
+                                        <exclude>META-INF/LICENSE</exclude>
+                                        <exclude>META-INF/NOTICE</exclude>
+                                        <exclude>META-INF/services/**</exclude>
+                                        <exclude>META-INF/maven/com.fasterxml.jackson.core/**</exclude>
```
rhauch, line 63: Does the existing `htrace-core4` library really contain Jackson classes that aren't in a shaded package?

htrace-core4-shaded/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+                            <filters>
+                                <filter>
+                                    <artifact>org.apache.htrace:htrace-core4</artifact>
+                                    <includes>
+                                        <include>**</include>
+                                    </includes>
+                                    <excludes>
+                                        <exclude>META-INF/MANIFEST.MF</exclude>
+                                        <exclude>META-INF/LICENSE</exclude>
+                                        <exclude>META-INF/NOTICE</exclude>
```
rhauch, line 61: Hmm, does the shading process add these back? If not, are we losing the license and notice for `htrace-core4`?

rhauch (commented):
It might be worth noting on this PR and probably the commit (if not a comment in the new shading POM file) that both `htrace-core4` and Jackson are licensed under ASL 2.0, and therefore shading does not affect licensing (tho we need to confirm this).

htrace-core4-shaded/pom.xml
kkonstantine: Good question. When there are multiple, which ones gets written is arbitrary. I changed the pom to keep the meta files from htrace and not jackson.

htrace-core4-shaded/pom.xml
kkonstantine: The version being removed is 2.4.0. I've run the integration tests for the various affected connectors with the produced package and there are no issues (removing this dependency breaks the tests).

kkonstantine: A similar approach has been discussed in: https://issues.apache.org/jira/browse/HDFS-15333
But no actual packages have been published.

kkonstantine:
Thanks for the reviews @rhauch. I've answered in line to your comments. 
Regarding the license, I've added the ASL v2 in the pom of the new package

htrace-core4-shaded/pom.xml
rhauch: No concern. I just was caught offguard that we were removing unshaded Jackson-related artifacts, but didn't notice the `maven` directory in the path.

rhauch (approved):
LGTM. Thanks for figuring this out, @kkonstantine. Nice work!

kkonstantine:
Thanks @rhauch !
I've tested locally with HDFS and other connectors. But I'll test with a draft PR on github too before releasing storage-common.

PR #159, opened 2020-11-02
Reshade transitive dependency avatica with newer jackson to remove CVEs

kkonstantine:
## Problem
The HDFS connector (Hadoop v2) depends on a few transitive dependencies that are old and are shading a version of jackson that has CVEs associated with it.

## Solution
We are reshading this dependency by removing the old jackson version and replacing it with a newer one that does not have CVEs. HDFS connector can import this dependency and get a clean avatica 1.8.0 package (the connector is incompatible with newer avatica versions).

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


avatica-shaded/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,91 @@
...
+      <version>6.0.0</version>
+      <type>zip</type>
+      <classifier>resources</classifier>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+  <properties>
+    <avatica.version>1.8.0</avatica.version>
+    <jackson.version>2.10.5</jackson.version>
+  </properties>
```
rhauch, line 90: Seems strange to have these at the bottom.

avatica-shaded/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+                            <filters>
+                                <filter>
+                                    <artifact>org.apache.calcite.avatica:avatica</artifact>
+                                    <includes>
+                                        <include>**</include>
+                                    </includes>
+                                    <excludes>
+                                        <exclude>META-INF/MANIFEST.MF</exclude>
+                                        <exclude>META-INF/LICENSE</exclude>
+                                        <exclude>META-INF/NOTICE</exclude>
```
rhauch, line 61: Hmm, does the shading process add these back for avatica and Jackson? If not, are we losing the license and notice for `avatica`? Is there anything else shaded in avatica that will not be included in the generated files?

avatica-shaded/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+
+    <artifactId>kafka-connect-storage-common-avatica-shaded</artifactId>
+    <name>kafka-connect-storage-common-avatica-shaded</name>
+    
+    <description>avatica shaded package to replace jackson dependencies without CVEs</description>
+    <inceptionYear>2020</inceptionYear>
+
+    <properties>
+        <avatica.version>1.8.0</avatica.version>
+        <jackson.version>2.10.5</jackson.version>
```
rhauch, line 30: What Jackson version is shaded in the original library? And how will we gain confidence that there are no compatibility problems by replacing that version with 2.10.5?

Should this use a variable defined in the parent POM? I guess maybe not, since it doesn't have to be the same Jackson version used by the connector since this is shaded and used only by the avatica library.

avatica-shaded/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,91 @@
...
+                  <artifact>org.apache.calcite.avatica:avatica</artifact>
+                  <includes>
+                    <include>**</include>
+                  </includes>
+                  <excludes>
+                    <exclude>META-INF/MANIFEST.MF</exclude>
+                    <exclude>META-INF/LICENSE</exclude>
+                    <exclude>META-INF/NOTICE</exclude>
+                    <exclude>META-INF/services/com.fasterxml.jackson.core*</exclude>
+                    <exclude>META-INF/maven/com.fasterxml.jackson.core/**</exclude>
```
rhauch, line 42: Does the existing `avatica`library really contain Jackson classes that aren't in a shaded package?

avatica-shaded/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+                                    <artifact>org.apache.calcite.avatica:avatica</artifact>
+                                    <includes>
+                                        <include>**</include>
+                                    </includes>
+                                    <excludes>
+                                        <exclude>META-INF/MANIFEST.MF</exclude>
+                                        <exclude>META-INF/LICENSE</exclude>
+                                        <exclude>META-INF/NOTICE</exclude>
+                                        <exclude>META-INF/services/com.fasterxml.jackson.core*</exclude>
+                                        <exclude>META-INF/maven/com.fasterxml.jackson.core/**</exclude>
```
rhauch, line 63: Does the existing `avatica`library really contain Jackson classes that aren't in a shaded package?

avatica-shaded/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,91 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <parent>
+    <artifactId>kafka-connect-storage-common-parent</artifactId>
+    <groupId>io.confluent</groupId>
+    <version>10.0.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>kafka-connect-storage-common-avatica-shaded</artifactId>
+  <name>kafka-connect-storage-common-avatica-shaded</name>
+  <description>avatica shaded package to replace jackson dependencies without CVEs</description>
```
rhauch, line 11: Can you explain how this file is used? It has the same artifactId as the `avatica-shaded/pom.xml` file, and it's not explicitly referenced as far as I can see.

rhauch (commented):
It might be worth noting on this PR and probably the commit (if not a comment in the new shading POM file) that both `avatica` and Jackson are licensed under ASL 2.0, and therefore shading does not affect licensing (tho we need to confirm this).

avatica-shaded/dependency-reduced-pom.xml
kkonstantine: This files was added unintentionally. It's produced in every run. I thought of adding it to `.gitignore`

avatica-shaded/pom.xml
kkonstantine: We were adding jackson. But better to keep the original here too.

avatica-shaded/pom.xml
kkonstantine: I was thinking that we won't like to upgrade the shaded jackson as frequently, so pin here made sense. This is a recent clear version that seems to work. Trying earlier ones would probably bring in other CVEs.

avatica-shaded/dependency-reduced-pom.xml
kkonstantine: I'm removing all the meta files that contain jackson from the package. These were in the original package.

avatica-shaded/pom.xml
kkonstantine: These directories contained only pom files. 
`pom.properties` and `pom.xml`. They are included when the new dependencies are shaded in. 
The `services` contains service loader configuration files and are also correctly readded by the newer dependencies.

avatica-shaded/dependency-reduced-pom.xml
kkonstantine: Removed. Added by mistake.

kkonstantine:
Thanks for the reviews @rhauch. I've answered in line. 
Regarding the license, I've added the ASL v2 in the pom of the new package

rhauch (approved):
LGTM. Thanks for figuring this out, @kkonstantine. Nice work!

kkonstantine:
Thanks! Rebased to resolve conflicts.

Issue #126, opened 2020-03-05
Improve to storage partitioning scheme to be flexible and configurable at runtime.


hariprasad-k:
Currently, the partitioning scheme includes topic name by default in all storage partitioning scheme. It will be flexible to have it configurable and let users chose. 

Along the same lines, the partitioning scheme also assumes the 'timestamp' column unit in seconds by default. Instead, it will be flexible to have this configurable and let users specify a scale factor to boost or shrink to seconds scale to support units in different scales.
This is similar to https://github.com/confluentinc/kafka-connect-storage-common/pull/122, but a little more generic.

cricket007:
If you didn't have the topic name, how would you separate out directories by topic partition? What exactly is your use case that cannot be solved by a regex router or a custom partitioner that you add to your own classpath?

hariprasad-k:
> If you didn't have the topic name, how would you separate out directories by topic partition? 

This improvement makes it more flexible for users to specify a full output path, that already consists topic name included with different delimiters or directory hierarchies based on topic names. 

> What exactly is your use case that cannot be solved by a regex router or a custom partitioner that you add to your own classpath?

For us, this is particularly interesting to facilitate smoother migration track (from Camus to Kafka Connect) and have all HDFS output paths (along with Hive tables) fully backward compatible as before. 

And further, if we use a regex router or a custom partitioner, this will break EOS semantics (offset recovery) achieved by  HDFS file naming conventions. Hence, we propose this change along with this PR https://github.com/confluentinc/kafka-connect-hdfs/pull/516 to retain the EOS semantics.

OneCricketeer:
> already consists topic name included with different delimiters or directory hierarchies based on topic names

So, you've created some other system to maintain this directory structure? 

> migration track (from Camus to Kafka Connect)  

I've not heard of Kafka Connect being an exact migration path (used to use Camus as well). Apache Gobblin is the next iteration of Camus

hariprasad-k:
> > already consists topic name included with different delimiters or directory hierarchies based on topic names
> 
> So, you've created some other system to maintain this directory structure?

We have external Hive tables based on this HDFS directory structure. So, with this patch, we were able to retain same directory structure..

> > migration track (from Camus to Kafka Connect)
> 
> I've not heard of Kafka Connect being an exact migration path (used to use Camus as well). Apache Gobblin is the next iteration of Camus

[...]

PR #160, opened 2020-11-03
 Improvement to have specify scaling factor for timestamp column used in partitioning by configuration

hariprasad-k:
## Problem
#126 Improve to storage partitioning scheme to be flexible and configurable at runtime

## Solution
This provides ability to specify scaling factor for timestamp column used in partitioning by configuration

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


hariprasad-k:
The patch also fixes #121 too

PR #162, opened 2020-11-04
CC-1360: Enable enhanced avro schema support by default in storage sink connectors

kkonstantine:
## Problem
Changing the default required the release of a major version

## Solution
Switch the default to enhanced avro schema support

- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Only 10.0.x and beyond

### DIFF
--- a/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
+++ b/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
@@ -101,7 +101,7 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
   public static final String SCHEMA_CACHE_SIZE_DISPLAY = "Schema Cache Size";
 
   public static final String ENHANCED_AVRO_SCHEMA_SUPPORT_CONFIG = "enhanced.avro.schema.support";
-  public static final boolean ENHANCED_AVRO_SCHEMA_SUPPORT_DEFAULT = false;
+  public static final boolean ENHANCED_AVRO_SCHEMA_SUPPORT_DEFAULT = true;
   public static final String ENHANCED_AVRO_SCHEMA_SUPPORT_DOC =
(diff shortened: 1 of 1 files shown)


dosvath (commented):
Before moving forward, I have a few questions:

- It looks like Enhanced Avro Schema support appears as `false` default in multiple of our docs. [Schema registry docs](https://docs.confluent.io/current/schema-registry/connect.html), [S3 Sink docs](https://docs.confluent.io/current/connect/kafka-connect-s3/configuration_options.html). 
- We should also make sure to consider issues [like this enum handling issue](https://github.com/confluentinc/schema-registry/issues/658) when using enhanced schema support to avoid possible problems with AvroData in our sink connectors. 
- What is the reasoning behind turning this default to `true`?

kkonstantine:
Thanks for the review @dosvath 

> * It looks like Enhanced Avro Schema support appears as `false` default in multiple of our docs. [Schema registry docs](https://docs.confluent.io/current/schema-registry/connect.html), [S3 Sink docs](https://docs.confluent.io/current/connect/kafka-connect-s3/configuration_options.html).

The docs are in a separate repo.

> * We should also make sure to consider issues [like this enum handling issue](https://github.com/confluentinc/schema-registry/issues/658) when using enhanced schema support to avoid possible problems with AvroData in our sink connectors.

We are not removing the config here. We are just turning it on by default. Users have still the ability to turn it off. 

> * What is the reasoning behind turning this default to `true`? 

This config has been introduced a while back. It's potentially an incompatible change for users, so it had to wait for a major version release to be switched from `false` to `true`. But enabling it, is the mainstream behavior we want to have. 

> In general I feel cautious about enabling a feature by default that used to be off due to the possibility of breaking changes with users' existing connector configurations.

That's why we waited for a major release.

dosvath (approved):
Thanks for the answers, makes sense 👍

dosvath:
LGTM

PR #168, opened 2020-12-07
CC-13001: Remove avro-converter dependency and add avro-data dependency

Prathibha-m:
## Problem
The current cc-docker image has kafka-connect-avro-converter in multiple places:

/usr/share/java/connectors/plugins/converters

/opt/confluent/libs

S3 sink, ADLS2, GCP functions, Azure functions, Dataproc sink, Azure blob storage, GCS sink connectors.

It’s problematic to have the same library in multiple locations, and we need to worry about version consistency. We want to have this library in /usr/share/java/connectors/plugins/converters only, and treat it as another plugin.

## Solution
Now that we have the `avro-data` dependency available, remove `avro-converter` dependency from downstream blob connectors https://confluentinc.atlassian.net/browse/CC-7169

- [x] yes ADLS Gen2 Sink, Azure Blob Sink, GCS Sink, S3 Sink
- [ ] no

##### Testing done:

Tested with IT's in Azure Blob Storage https://github.com/confluentinc/kafka-connect-azure-blob-storage/pull/125
- [ ] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -36,7 +36,7 @@
         </dependency>
         <dependency>
             <groupId>io.confluent</groupId>
-            <artifactId>kafka-connect-avro-converter</artifactId>
+            <artifactId>kafka-connect-avro-data</artifactId>
         </dependency>
         <dependency>
             <groupId>joda-time</groupId>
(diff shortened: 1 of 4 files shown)


levzem:
@Prathibha-m we should not be merging to `post` branches also,
 would be good to run an IT on one of the storage connectors to make sure this change doesn't break anything

levzem (approved):
LGTM thanks @Prathibha-m !

PR #169, opened 2020-12-28
MINOR: pin jamon dependency to fix build errors

levzem:

## Problem
jamon runtime is causing build errors for some storage connectors i.e [HDFS](https://github.com/confluentinc/kafka-connect-hdfs/tree/10.0.x)

due to https://www.jfrog.com/jira/browse/RTFACT-16033, suppressing consistency checks is not an option

## Solution
pin this dependency and release a new version of storage common

- [ ] yes
- [x] no

## Test Strategy
pinned the snapshot version to HDFS and made sure it built and ran all of the hive tests correctly

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
release `v10.0.2` for the storage connectors to use

### DIFF
--- a/hive/pom.xml
+++ b/hive/pom.xml
@@ -95,6 +95,10 @@
                     <groupId>log4j</groupId>
                     <artifactId>log4j</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.jamon</groupId>
+                    <artifactId>jamon-runtime</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
         <!-- Explicitly include working version of xml-apis -->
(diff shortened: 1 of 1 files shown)


hive/pom.xml
```diff
@@ -134,6 +138,11 @@
             <artifactId>avro-mapred</artifactId>
             <version>${avro.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.jamon</groupId>
+            <artifactId>jamon-runtime</artifactId>
+            <version>2.4.1</version>
```
kkonstantine, line 144: please add the version as a property

kkonstantine (commented):
Left a comment. 
Related to that is that the description is quite generic. It would be good to know the exact root cause if we've identified it.

levzem:
> Left a comment.
> Related to that is that the description is quite generic. It would be good to know the exact root cause if we've identified it.

@kkonstantine I have updated the description. The reason is due to https://www.jfrog.com/jira/browse/RTFACT-16033

kkonstantine:
Thanks @levzem 
I'm having trouble building on 10.0.x as well. Should we retarget the PR?

levzem:
@kkonstantine good catch, `10.0.x` was my intention

kkonstantine (approved):
Thanks
LGTM, if it builds.

PR #171, opened 2021-01-04
CC-13056: address http and jackson CVEs

levzem:

## Problem
there are still multiple CVEs (jackson-databind and http-components)

## Solution
bump common parent and the CVEs versions to clean ones (twistlock shows clean)

- [ ] yes
- [x] no

## Test Strategy
bug fix version bumps so harmless
built a local version and tried it downstream with HDFS and everything passed

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
`10.0.x` where the CVEs are reported

### DIFF
--- a/avatica-shaded/pom.xml
+++ b/avatica-shaded/pom.xml
@@ -51,6 +51,7 @@ language governing permissions and limitations under the License. -->
                                 <includes>
                                     <include>org.apache.calcite.avatica:*</include>
                                     <include>com.fasterxml.jackson.core:*</include>
+                                    <include>org.apache.httpcomponents</include>
                                 </includes>
                             </artifactSet>
                             <filters>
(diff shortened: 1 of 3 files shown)


kkonstantine (commented):
Looks good. But have you checked the shaded packages too (avatica and htrace)?
We'll need to clean those as well. Let's check that jackson databind is on right version there before we release. 
E.g. https://github.com/confluentinc/kafka-connect-storage-common/blob/master/avatica-shaded/pom.xml#L112

levzem:
> Looks good. But have you checked the shaded packages too (avatica and htrace)?
> We might need to clean those as well. Let's check that jackson databind is on right version there before we release.
> E.g. https://github.com/confluentinc/kafka-connect-storage-common/blob/master/avatica-shaded/pom.xml#L112

good catch @kkonstantine, updated

kkonstantine (approved):
Looks clean. 
LGTM if things don't break downstream.

PR #172, opened 2021-01-12
CC-13532: Pin storage common version in parent

dosvath:
## Problem
Currently the parent pom uses `${project.version}` for the submodules `kafka-connect-storage-common`, `kafka-connect-storage-core`, `kafka-connect-storage-format`, `kafka-connect-storage-partitioner`, `kafka-connect-storage-wal`. 

This becomes an issue when building connectors that use this parent, as the `${project.version}` in a parent pom refers to the version of the child, not the parent. The value gets re-interpreted by maven every time on a build and refers to the version of the project built at that moment, not the version of the parent. This has resulted in connector build job errors as `SNAPSHOT` dependencies are not allowed. 

```
01:01:13  [ERROR] Failed to execute goal org.apache.maven.plugins:maven-release-plugin:2.5.3:prepare (default-cli) on project kafka-connect-storage-cloud: 
Can't release project due to non released dependencies :
01:01:13  [ERROR]     io.confluent:kafka-connect-storage-common:jar:10.0.0-SNAPSHOT:compile
01:01:13  [ERROR]     io.confluent:kafka-connect-storage-core:jar:10.0.0-SNAPSHOT:compile
01:01:13  [ERROR]     io.confluent:kafka-connect-storage-format:jar:10.0.0-SNAPSHOT:compile
01:01:13  [ERROR]     io.confluent:kafka-connect-storage-partitioner:jar:10.0.0-SNAPSHOT:compile
01:01:13  [ERROR] in project 'kafka-connect-storage-cloud' (io.confluent:kafka-connect-storage-cloud:pom:10.0.0-SNAPSHOT)
```
[In this example](https://github.com/confluentinc/kafka-connect-hdfs/pull/529/files) we also had to replace the `SNAPSHOT` version with a non-snapshot one for the build job to pass. 
## Solution

[Current workaround](https://github.com/confluentinc/kafka-connect-hdfs/commit/5ce49b67379ab63b548b650a4a0908ad547af6ec) defines the version in the connector pom. Instead of overwriting this version property in each child repo, we should define the version in the parent. This will also allow us to remove the version pinned in the inherited connector poms and remove the `<version>` tags of the dependencies there for a cleaner configuration.

- [ ] yes
- [X] no

##### Testing done:
- [X] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Backporting to `10.0.x` as CP connectors were de-coupled starting that version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,7 @@
     </modules>
 
     <properties>
+        <kafka.connect.storage.common.version>10.0.2</kafka.connect.storage.common.version>
         <commons-io.version>2.4</commons-io.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
         <hadoop.version>2.10.1</hadoop.version>
@@ -94,27 +95,27 @@
             <dependency>
                 <groupId>io.confluent</groupId>
                 <artifactId>kafka-connect-storage-common</artifactId>
-                <version>${project.version}</version>
+                <version>${kafka.connect.storage.common.version}</version>
             </dependency>
             <dependency>
                 <groupId>io.confluent</groupId>
                 <artifactId>kafka-connect-storage-core</artifactId>
-                <version>${project.version}</version>
+                <version>${kafka.connect.storage.common.version}</version>
             </dependency>
             <dependency>
                 <groupId>io.confluent</groupId>
                 <artifactId>kafka-connect-storage-format</artifactId>
-                <version>${project.version}</version>
+                <version>${kafka.connect.storage.common.version}</version>
             </dependency>
             <dependency>
                 <groupId>io.confluent</groupId>
                 <artifactId>kafka-connect-storage-partitioner</artifactId>
-                <version>${project.version}</version>
+                <version>${kafka.connect.storage.common.version}</version>
             </dependency>
             <dependency>
                 <groupId>io.confluent</groupId>
                 <artifactId>kafka-connect-storage-wal</artifactId>
-                <version>${project.version}</version>
+                <version>${kafka.connect.storage.common.version}</version>
             </dependency>
             <dependency>
                 <groupId>io.confluent</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
     </modules>
 
     <properties>
-        <kafka.connect.storage.common.version>10.0.2</kafka.connect.storage.common.version>
+        <kafka.connect.storage.common.version>10.0.3-SNAPSHOT</kafka.connect.storage.common.version>
         <commons-io.version>2.4</commons-io.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
         <hadoop.version>2.10.1</hadoop.version>


C0urante:
This looks good but we should verify that the `kafka.connect.storage.common.version` gets bumped during the release job. If not, might have to issue a follow-up PR there as well before merging this.

dosvath:
> This looks good but we should verify that the `kafka.connect.storage.common.version` property gets bumped during the release job. If not, might have to issue a follow-up PR there as well before merging this. Or, we could hardcode the versions in the dependencies here to match the version of the project (i.e., `10.0.3-SNAPSHOT`), which I believe should work with the release job in its current state as that's how we [used to do things](https://github.com/confluentinc/connect-plugins-common/blob/v0.4.0/pom.xml#L84-L88) with `connect-plugins-common`.

Thanks @C0urante, would your second suggestion also require configuring the build job to bump those hardcoded ones? 
For using the project version, ex  `10.0.3-SNAPSHOT` the concern would be the issue I mentioned in the problem section of the description.

C0urante:
@dosvath I _believe_ (it's been a while) that if we hardcode the versions in the POM like this:
```xml
<version>10.0.3-SNAPSHOT</version>
...
<build>
  <dependencyManagement>
    <dependencies>
      <dependency>
        <artifactId>kafka-connect-storage-common</artifactId>
        <version>10.0.3-SNAPSHOT</version>
      </dependency>
    </dependencies>
  </dependencyManagement>
</build>
```

that, for the officially-released artifact, the POM will be transformed to look like this (note that the project version _and_ the dependency version have been updated):

```xml
<version>10.0.3</version>
...
<build>
  <dependencyManagement>
    <dependencies>
      <dependency>
        <artifactId>kafka-connect-storage-common</artifactId>
        <version>10.0.3</version>
      </dependency>
    </dependencies>
  </dependencyManagement>
</build>```

and we there shouldn't be any issues with snapshot dependencies.

File: pom.xml [hunk 1]
```diff
@@ -65,6 +65,7 @@
     </modules>
 
     <properties>
+        <kafka.connect.storage.common.version>10.0.2</kafka.connect.storage.common.version>
```
kkonstantine, line 68: Essentially what you need is: 
```suggestion
        <kafka.connect.storage.common.version>${project.version}</kafka.connect.storage.common.version>
```

See if that works in the properties section.

kkonstantine: Hmm, scratch that. It won't work. I remembered. 
I checked the `mvn help:effective-pom`

C0urante: I changed this to match the project version (`10.0.3-SNAPSHOT`) and then ran a release dry run on it; the property was automatically updated to `10.0.3` for the tagged release, and `10.0.4-SNAPSHOT` for the commit immediately after.

I think all that's necessary for this to work is to change this line to use `10.0.3-SNAPSHOT` as the version, and it'll be automatically updated every time we do a release.

dosvath: I think that is our best path here +1

C0urante (approved):
LGTM

PR #174, opened 2021-01-29
Exclude commons-codec from the Avatica shaded JAR and include its newer version separately to resolve CVE

avocader:
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/avatica-shaded/pom.xml
+++ b/avatica-shaded/pom.xml
@@ -22,7 +22,7 @@ language governing permissions and limitations under the License. -->
     <artifactId>kafka-connect-storage-common-avatica-shaded</artifactId>
     <packaging>jar</packaging>
     <name>kafka-connect-storage-common-avatica-shaded</name>
-    
+
     <description>avatica shaded package to replace jackson dependencies without CVEs</description>
     <inceptionYear>2020</inceptionYear>
 
@@ -52,6 +52,7 @@ language governing permissions and limitations under the License. -->
                                     <include>org.apache.calcite.avatica:*</include>
                                     <include>com.fasterxml.jackson.core:*</include>
                                     <include>org.apache.httpcomponents</include>
+                                    <include>commons-codec</include>
                                 </includes>
                             </artifactSet>
                             <filters>
@@ -66,6 +67,8 @@ language governing permissions and limitations under the License. -->
                                         <exclude>com/fasterxml/jackson/**</exclude>
                                         <exclude>META-INF/maven/org.apache.httpcomponents/**</exclude>
                                         <exclude>org/apache/calcite/avatica/org/apache/http/**</exclude>
+                                        <exclude>org/apache/calcite/avatica/org/apache/commons/codec/**</exclude>
+                                        <exclude>META-INF/maven/commons-codec/commons-codec/**</exclude>
                                     </excludes>
                                 </filter>
                                 <filter>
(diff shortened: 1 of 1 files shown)


kkonstantine (commented):
Let's target 10.0.x since this seems like it can be backported

avatica-shaded/pom.xml
```diff
@@ -130,6 +130,17 @@ language governing permissions and limitations under the License. -->
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpclient</artifactId>
             <version>${httpclient.version}</version>
+            <exclusions>
```
kkonstantine, line 133: is commons codec part of httpclient? 
This might not remove it and you might need to filter it as the three above (and then re-add it here)

kkonstantine: You're right. It's a transitive of httpclient

kkonstantine: Actually I take that back. 
`./org/apache/calcite/avatica/org/apache/commons/codec` is shaded too. 
So, needs to be filtered as above (including the meta-inf files) and to be added as an explicit dependency

avocader: Yes, I see it's coming from `httpclient`:
```
mvn dependency:tree -Dverbose
...
[INFO] +- org.apache.httpcomponents:httpclient:jar:4.5.13:compile
[INFO] |  +- commons-logging:commons-logging:jar:1.2:compile
[INFO] |  \- commons-codec:commons-codec:jar:1.15:compile
...
```
But that didn't work indeed.

avocader: Actually, after looking at the content of the shaded JAR - it doesn't have the classes of this JAR, only maven's metadata:
```
~/workspace/dev/kafka-connect-storage-common/avatica-shaded/target> unzip kafka-connect-storage-common-avatica-shaded-10.1.0-SNAPSHOT.jar -d kafka-connect-storage-common-avatica-shaded-10.1.0-SNAPSHOT
...
find kafka-connect-storage-common-avatica-shaded-10.1.0-SNAPSHOT -name "commons-codec"
kafka-connect-storage-common-avatica-shaded-10.1.0-SNAPSHOT/META-INF/maven/commons-codec
kafka-connect-storage-common-avatica-shaded-10.1.0-SNAPSHOT/META-INF/maven/commons-codec/commons-codec
...
tree kafka-connect-storage-common-avatica-shaded-10.1.0-SNAPSHOT/META-INF/maven/commons-codec
kafka-connect-storage-common-avatica-shaded-10.1.0-SNAPSHOT/META-INF/maven/commons-codec
└── commons-codec
    ├── pom.properties
    └── pom.xml
```

So I just excluded the metadata from the shaded JAR.

kkonstantine: As we discussed offline, the files are there, shaded in the jar under:
`org/apache/calcite/avatica/org/apache/commons/codec`
We'll have to filter out those too

avocader:
@kkonstantine with the latest changes, the final JAR has only one set of classes from `commons-codec`, uses the `1.15` version.:
```
cd kafka-connect-storage-common-avatica-shaded-10.0.4-SNAPSHOT
...
find . -name "BinaryDecoder.class"
./org/apache/commons/codec/BinaryDecoder.class
...
cat META-INF/maven/commons-codec/commons-codec/pom.properties|grep version
version=1.15
```

kkonstantine (approved):
LGTM

kkonstantine:
Also, let's make sure that in the end the title corresponds to the actual changes we add

PR #175, opened 2021-01-30
Update pom to exclude org/apache/http from the shaded JAR.

avocader:
## Problem

Looks like we have two sets of classes in `org.apache.http` package: one of the correct version(`4.5.13`) and one of the old version(`4.5.2`).

```
mvn clean package
...
unzip kafka-connect-storage-common-avatica-shaded-10.0.4-SNAPSHOT.jar -d kafka-connect-storage-common-avatica-shaded-10.0.4-SNAPSHOT
...
cd kafka-connect-storage-common-avatica-shaded-10.0.4-SNAPSHOT
...
find . -name "HttpClient.class"
./org/apache/calcite/avatica/org/apache/http/client/HttpClient.class
./org/apache/http/client/HttpClient.class
...
cat org/apache/http/client/version.properties|grep release
info.release   = 4.5.13
...
cat org/apache/calcite/avatica/org/apache/http/client/version.properties|grep release
info.release   = 4.5.2
```

## Solution
Exclude the `org/apache/httpcomponents/**` package from the shaded JAR.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/avatica-shaded/pom.xml
+++ b/avatica-shaded/pom.xml
@@ -64,9 +64,8 @@ language governing permissions and limitations under the License. -->
                                         <exclude>META-INF/services/com.fasterxml.jackson.core*</exclude>
                                         <exclude>META-INF/maven/com.fasterxml.jackson.core/**</exclude>
                                         <exclude>com/fasterxml/jackson/**</exclude>
-                                        <exclude>META-INF/services/org.apache.httpcomponents*</exclude>
                                         <exclude>META-INF/maven/org.apache.httpcomponents/**</exclude>
-                                        <exclude>org/apache/httpcomponents/**</exclude>
+                                        <exclude>org/apache/calcite/avatica/org/apache/http/**</exclude>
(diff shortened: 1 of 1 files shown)


avatica-shaded/pom.xml
```diff
@@ -64,9 +64,8 @@ language governing permissions and limitations under the License. -->
                                         <exclude>META-INF/services/com.fasterxml.jackson.core*</exclude>
                                         <exclude>META-INF/maven/com.fasterxml.jackson.core/**</exclude>
                                         <exclude>com/fasterxml/jackson/**</exclude>
-                                        <exclude>META-INF/services/org.apache.httpcomponents*</exclude>
```
avocader, line 67: This is a no-op.

kkonstantine (approved):
LGTM it this still applies here

PR #177, opened 2021-03-03
MINOR: Use PartitionException instead of ConnectException when unable to extract the record timestamp in TimeBasedPartitioner

dosvath:
## Problem
Throwing `ConnectException` but failure is caused by malformed data. 

## Solution
Revise to `DataException` for a more appropriate exception type.

- [ ] yes
- [X] no

##### Testing done:
- [X] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Targeting `10.0.x`


kkonstantine (approved):
LGTM on green build. Thanks @dosvath

kkonstantine (approved):

PR #180, opened 2021-03-13
Handle Field partitioner to pick value from Map structure

pushpavanthar:
## Problem
Field Partitioner could not pick value if the incoming record is of the form of map

## Solution
Handled to fetch attribute value when the payload of Map type

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


pushpavanthar (commented):
LGTM

PR #185, opened 2021-05-11
CC-14522: resolve hadoop, jackson, thrift, log4j, commons-collection, bean-utils, netty, zookeeper, commons-codec, and httpclient CVEs

levzem:

## Problem
lots of CVEs leftover in 5.2.x and on

## Solution
resolve them using various methods
https://twistlock.tools.confluent-internal.io/#!/monitor/vulnerabilities/images/ci?search=storage-common:PR-185
all CVEs are resolved, the ones left are in S3 and HDFS
except for guava which i was unable to resolve because there is a backwards incompatible change that HDFS relies on
some CVEs have no resolution i.e `commons-collection`

- [x] yes
- [ ] no

##### If yes, where?
other CP connectors (S3 + HDFS)

## Test Strategy
tested against HDFS connector locally

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport to `5.2.x`

### DIFF
--- a/format/pom.xml
+++ b/format/pom.xml
@@ -29,6 +29,12 @@
     <packaging>jar</packaging>
     <name>kafka-connect-storage-format</name>
 
+    <properties>
+        <jackson.groupId>com.fasterxml.jackson.core</jackson.groupId>
+        <jackson.package>com.fasterxml.jackson</jackson.package>
+        <shade.prefix>shaded.parquet</shade.prefix>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>io.confluent</groupId>
@@ -45,6 +51,12 @@
         <dependency>
             <groupId>org.apache.parquet</groupId>
             <artifactId>parquet-avro</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.parquet</groupId>
+                    <artifactId>parquet-jackson</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>junit</groupId>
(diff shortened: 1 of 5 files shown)


kkonstantine:
Last scan still shows a number of CVE unresolved. jackson databind 2.4.0 in the shaded jars as well as jackson asl and a few more. Have we confirmed that the fixes are ported accurately from recent CP versions?

levzem:
@kkonstantine 

as mentioned above all the CVEs that are left are being brought in by S3 and HDFS, for some reason they are being displayed on `storage-common`, see below for an example

Screen Shot 2021-05-13 at 10 34 09 AM

a few other CVEs we are unable to fix such `asl`, 'guava', and `commons-collections` as no fix versions exist and are required by the connector to function.

format/pom.xml
```diff
@@ -45,11 +51,59 @@
...
+                            <minimizeJar>false</minimizeJar>
+                            <createSourcesJar>false</createSourcesJar>
+                            <artifactSet>
+                                <includes>
+                                    <include>${jackson.groupId}:*</include>
+                                </includes>
+                            </artifactSet>
+                            <filters>
+                                <filter>
+                                    <artifact>${jackson.groupId}:*</artifact>
```
ilanjiR, line 91: Why are we filtering `${jackson.groupId}:*` if we're explicitly adding it above ^ ?

pom.xml
```diff
@@ -60,21 +60,34 @@
...
+        <hadoop.version>2.10.1</hadoop.version>
         <hive.version>1.2.2</hive.version>
-        <joda.version>2.9.6</joda.version>
-        <licenses.version>5.2.5-SNAPSHOT</licenses.version>
-        <parquet.version>1.8.2</parquet.version>
         <httpclient.version>4.5.13</httpclient.version>
         <httpcore.version>4.4.4</httpcore.version>
+        <jackson.version>2.10.5</jackson.version>
+        <jackson.databind.version>2.10.5.1</jackson.databind.version>
+        <jackson.mapper.asl.version>1.9.13</jackson.mapper.asl.version>
```
ilanjiR, line 81: I'm guessing there's no universal version for these 3 modules that addresses all the CVEs

ilanjiR (commented):
Couple of questions, but looks good otherwise

pom.xml
levzem: correct

format/pom.xml
kkonstantine: We need to first remove the old version shaded in the package and then reshade it with the version we supply here.

ilanjiR (approved):
LGTM

PR #186, opened 2021-05-13
Upgrade jackson-mapper-asl to 1.9.14.jdk17-redhat-00001 and fix CVE

ilanjiR:
## Problem
Critical CVE for jackson-mapper

## Solution
Upgrade to redhat version which has no CVE

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
         <httpcore.version>4.4.4</httpcore.version>
         <jackson.version>2.10.5</jackson.version>
         <jackson.databind.version>2.10.5.1</jackson.databind.version>
-        <jackson.mapper.asl.version>1.9.13</jackson.mapper.asl.version>
+        <jackson.mapper.asl.version>1.9.14.jdk17-redhat-00001</jackson.mapper.asl.version>
         <jetty.version>9.4.40.v20210413</jetty.version>
         <jline.version>2.12.1</jline.version>
         <joda.version>2.9.6</joda.version>
@@ -96,6 +96,10 @@
             <id>confluent</id>
             <name>Confluent</name>
             <url>${confluent.maven.repo}</url>
+        </repository>
+        <repository>
+            <id>redhat-repo</id>
+            <url>https://maven.repository.redhat.com/ga/</url>
         </repository>
     </repositories>
 


core/pom.xml
```diff
@@ -37,6 +37,12 @@
         <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>kafka-connect-avro-data</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.glassfish.jersey.core</groupId>
+                    <artifactId>jersey-common</artifactId>
```
kkonstantine, line 43: Seems like something that would be solved in the avro-data package. Also, is the removal safe?

ilanjiR: All the tests passed for the repo. Is there any other way to check if it's a safe change?

pom.xml
```diff
@@ -133,7 +133,8 @@
             <dependency>
                 <groupId>io.confluent</groupId>
                 <artifactId>kafka-connect-avro-data</artifactId>
-                <version>${confluent.version}</version>
+                <version>${confluent.version}</version><exclusions>
+            </exclusions>
```
kkonstantine, line 141: this change is a leftover

core/pom.xml
kkonstantine: Not convinced we should pin this to a higher/different version. 
The most recent fix that sets the version in avro data is: https://github.com/confluentinc/rest-utils/pull/246
If that's what the rest of CP is using maybe it makes sense to stay in sync.

pom.xml
```diff
@@ -247,6 +251,11 @@
             <version>${confluent.log4j.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.glassfish.jersey.core</groupId>
+            <artifactId>jersey-common</artifactId>
```
kkonstantine, line 256: let's skip jersey which we'll decide later whether to upgrade to 2.34 or 3.0.2 and let's change the title to mention the important fix on mapper asl which will fix a breaking change in newer branches

hive/pom.xml
```diff
@@ -133,5 +133,10 @@
             <version>${junit.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>30.1.1-jre</version>
```
kkonstantine, line 139: Let's issue a separate PR in case we need to revert. 
Also let's get the version from common here: https://github.com/confluentinc/common/blob/5.5.x/pom.xml#L63

kkonstantine (commented):
LGTM on asl. Let's keep that change and issue separate PRs for the other two

levzem (commented):
did you test this with HDFS? otherwise lgtm but we should backport this to `5.2.x`

kkonstantine:
@ilanjiR please add the version to the title so that the release notes are easier to follow. 
Also, it still seems to me that we should upgrade and fix the issue in a single PR. But I might be missing the reason to issue two PRs. 
See comment here: https://github.com/confluentinc/kafka-connect-storage-common/pull/191#issuecomment-847440598

kkonstantine (approved):
LGTM

ilanjiR:
@levzem this was tested against hdfs 5.2.x and all the tests passed.

PR #189, opened 2021-05-14
Address CVE for guava by pinning to parent version

ilanjiR:
## Problem
Had a CVE for guava

## Solution
Inherit fixed version from common

- [ ] yes
- [x] no

## Test Strategy
Tested against S3 sink connector

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests


kkonstantine (commented):
@ilanjiR does this apply to older branches as well (e.g `5.2.x` ?)
Also, are there any side-effects in tests and system tests in connectors downstream that use this package?

PR #190, opened 2021-05-14
MINOR: exclude jetty-all to address CVE

levzem:

## Problem
`jetty-all` is a CVE that is getting pulled into HDFS

## Solution
exclude it

- [ ] yes
- [ ] no

## Test Strategy
tested against HDFS

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport to `5.2.x` where the CVE needs to be fixed

### DIFF
--- a/hive/pom.xml
+++ b/hive/pom.xml
@@ -79,6 +79,10 @@
                     <artifactId>hive-exec</artifactId>
                 </exclusion>
                 <exclusion>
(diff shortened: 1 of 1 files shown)


kkonstantine (approved):
LGTM if we ran tests on downstream connectors

PR #192, opened 2021-05-26
MINOR: Keep zookeeper as transitive dependency inherited from common parent pom

kkonstantine:
## Problem
When the versions of zk differ from Kafka tests and common, tests break

## Solution
Keep getting ZK as transitive dependency at the version that is defined in https://github.com/confluentinc/common/blob/master/pom.xml#L94

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests


kkonstantine:
Tested locally. Merging to fix the broken build in 5.5.x (IT tests). If needed will reiterate.

PR #200, opened 2021-06-03
Address libthrift CVEs by upgrading to 0.13.0

ilanjiR:
## Problem
Libthrift had a CVE

## Solution
Exclude libthrift

- [ ] yes
- [x] no

## Test Strategy
Tested against S3 sink

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests


pom.xml
```diff
@@ -86,7 +86,7 @@
         <licenses.version>5.2.5-SNAPSHOT</licenses.version>
         <netty.version>4.1.63.Final</netty.version>
         <parquet.version>1.11.0</parquet.version>
-        <thrift.version>0.9.3</thrift.version>
+        <thrift.version>0.9.3-1</thrift.version>
```
kkonstantine, line 89: This leaves 4 CVE open still. 
We need 0.13.0 which I believe works will will leave only one open

kkonstantine (approved):
I confirmed that tests pass for HDFS 2 and S3 connectors after this is merged in common (in 5.2.x)

LGTM

PR #203, opened 2021-07-02
CCMSG-1133: Decouple topic name from hive table name

erdody:
## Problem

We need to update the HiveUtil.createTable interface by adding the topic name so we can decouple the hive table name from the location path.

## Solution

Add an extra topic argument.

- [ ] yes
- [x] no

### DIFF
--- a/hive/src/main/java/io/confluent/connect/storage/hive/HiveUtil.java
+++ b/hive/src/main/java/io/confluent/connect/storage/hive/HiveUtil.java
@@ -28,18 +28,25 @@ import io.confluent.connect.storage.partitioner.Partitioner;
  */
 public abstract class HiveUtil {
 
-  protected String url;
+  protected final String url;
   protected final HiveMetaStore hiveMetaStore;
   protected final String delim;
 
   public HiveUtil(AbstractConfig connectorConfig, HiveMetaStore hiveMetaStore) {
-    this.url = connectorConfig.getString(StorageCommonConfig.STORE_URL_CONFIG);
+    this(connectorConfig,
+            hiveMetaStore,
+            connectorConfig.getString(StorageCommonConfig.STORE_URL_CONFIG));
+  }
+
+  public HiveUtil(AbstractConfig connectorConfig, HiveMetaStore hiveMetaStore, String url) {
     this.hiveMetaStore = hiveMetaStore;
     this.delim = connectorConfig.getString(StorageCommonConfig.DIRECTORY_DELIM_CONFIG);
+    this.url = url;
   }
 
   public abstract void createTable(String database, String tableName, Schema schema,
-                                   Partitioner<FieldSchema> partitioner);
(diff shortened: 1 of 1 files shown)


dosvath (approved):
LGTM

kkonstantine:
The change probably makes sense @erdody. I'd like to see how the other PR looks like too. Do you mind linking it here?
Also, it'd be nice to know what's our release plan for this change and the compatibility implications. If you can add this info to the PR template it'd be useful I think

erdody:
Thanks @kkonstantine , this is the other PR https://github.com/confluentinc/kafka-connect-hdfs/pull/574
This was asked by one specific client, so I was thinking this would be a new feature release (not sure if we need a new minor version), no need to patch older versions. 
In terms of compatibility, this would be an "internally" incompatible change (we need changes in the hdfs connector) but the connector changes should be externally backwards compatible (drop in replacement for clients).

kkonstantine:
Thanks. Yeah I was referring specifically to which version you suggest this to go in. A minor version makes sense I believe and we probably don't need a major version here. But wanted to hear what you think too. 
Of course, it's good to keep in mind that this is a public repo on its own and HDFS connector is the main downstream dependency, but might not be the only one (users might have their own implementations depending on storage-common).

erdody:
@kkonstantine I was thinking the latest one. @dosvath do you agree?

dosvath:
@erdody Yes I think `10.1.x` would make sense

kkonstantine:
That's why I wanted to discuss specific versions, because actually 10.1.x already exists. Targeting that branch means that we are introducing this change in a patch release. 

My understanding is that we want to merge in the `master` branch and then release `10.2.0` (after creating `10.2.x`)

hive/src/main/java/io/confluent/connect/storage/hive/HiveUtil.java
```diff
@@ -28,18 +28,25 @@
...
-  protected String url;
+  protected final String url;
   protected final HiveMetaStore hiveMetaStore;
   protected final String delim;
 
   public HiveUtil(AbstractConfig connectorConfig, HiveMetaStore hiveMetaStore) {
-    this.url = connectorConfig.getString(StorageCommonConfig.STORE_URL_CONFIG);
+    this(connectorConfig,
+            hiveMetaStore,
+            connectorConfig.getString(StorageCommonConfig.STORE_URL_CONFIG));
```
kkonstantine, line 38: alignment is a bit off. In connectors, tab space is equal to 2 spaces (as opposed to AK checkstyle). 
```suggestion
    this(
        connectorConfig,
        hiveMetaStore,
        connectorConfig.getString(StorageCommonConfig.STORE_URL_CONFIG)
    );
```
(did the editing on GH to give an example. Please make sure it's correct)

kkonstantine (approved):
One minor comment but otherwise LGTM

PR #210, opened 2021-07-30
CC-15305: Upgrade transitive dependency commons-compress to 1.21

kkonstantine:
## Problem
Outdated dependency

## Solution
pin version only in the hive module

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Applies to 5.4.x or newer

### DIFF
--- a/hive/pom.xml
+++ b/hive/pom.xml
@@ -123,11 +123,6 @@
             <version>${commons.lang3.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


kkonstantine:
Tested locally with both S3 and hdfs downstream repos.

dosvath (approved):
LGTM

PR #214, opened 2021-09-13
Extend config of RecordTimestampExtractor with date-time format pattern

gadomsky:
## Problem
Partitioning by String field is possible only if this field contains data in only one specific format. There is no easy way to override RecordFieldTimestampExtractor with custom datetime formatter. 
## Solution
This PR introduceses new configuration field with format pattern. 
Default value of this new configuration is empty String, cause there is no coressponding pattern to DatetimeFormatter beeing used currently - it is only possible to set this programmticaly.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests


OneCricketeer:
You should be able to use an SMT to convert the timestamp before it even reaches the partitioner

PR #220, opened 2021-11-18
Added int64 customer formatter

kerdosa:
## Problem

The avro `int64` type with logicalType `timestamp-millis` becomes Date class in Java. Current FieldPartitioner cast it to Number class which throws exception.

## Solution
If `int64` type is Date object, use the SimpleDateFormatter to convert the Date to string.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


kpatelatwork:
@kerdosa Thanks for the change, do you mind adding a unit and integration test for this?

PR #229, opened 2022-03-17
Declare protobuf-java in dependencyManagement to get version from common parent

BlueEldur:
## Problem
The version is defined upstream confluent/common but is needed downstream by users of this repo. It makes sense to define it here.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -175,6 +175,11 @@
                 <version>${jackson.databind.version}</version>
             </dependency>
             <dependency>
+                <groupId>com.google.protobuf</groupId>
+                <artifactId>protobuf-java</artifactId>
+                <version>${protobuf.version}</version>
+            </dependency>
+            <dependency>
                 <groupId>commons-beanutils</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -174,6 +174,11 @@
                 <artifactId>jackson-databind</artifactId>
                 <version>${jackson.databind.version}</version>
             </dependency>
+            <dependency>
+                <groupId>com.google.protobuf</groupId>
+                <artifactId>protobuf-java</artifactId>
+                <version>${protobuf.version}</version>
```
cjolivier01, line 180: Nice!  Probably, if necessary, should also do protobuf-java-util since some places import that one too and might as well keep them the same version.

BlueEldur: good catch. updated

kpatelatwork (approved):
LGTM and Thanks a lot  @BlueEldur

PR #232, opened 2022-03-22
Bump jackson versions to 2.13.2 to address cve's

BlueEldur:
## Problem
Minor version update(backwrads compatible) to 2.13.2.
Add related jackson dependencies to dependencymanagement to guard against jackson version mismatches.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
         <httpclient.version>4.5.13</httpclient.version>
         <httpcore.version>4.4.4</httpcore.version>
         <jackson.version>2.13.2</jackson.version>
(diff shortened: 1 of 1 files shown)


cjolivier01 (approved):
I don;t feel very strongly about it, but any reason not to merge these two properties?

BlueEldur:
I was thinking of the same, but I thought perhaps there may be some downstream dependencies that were directly using either of them @cjolivier01

BlueEldur:
actually I realize we can just have one reference the other. Woops. Made that change. @cjolivier01

PR #237, opened 2022-05-17
CCMSG-1906: Adress CVEs for ant, jsp, groovy, derby, netty and orc. 

snehashisp:
## Problem
List of CVE that this PR addresses. 
https://confluentinc.atlassian.net/browse/CCMSG-1900
https://confluentinc.atlassian.net/browse/CCMSG-1894
https://confluentinc.atlassian.net/browse/CCMSG-1896
https://confluentinc.atlassian.net/browse/CCMSG-1901
https://confluentinc.atlassian.net/browse/CCMSG-1904
https://confluentinc.atlassian.net/browse/CCMSG-1908
https://confluentinc.atlassian.net/browse/CCMSG-1910
https://confluentinc.atlassian.net/browse/CCMSG-1912

## Solution
Since all of these dependencies are brought in through hive we have to override those dependencies through dependency management. Some of the dependencies can be removed.

- [ ] yes
- [ ] no

## Test Strategy
Tested this locally by installing the maven artefact and using it in hdfs sink connector. All test tests pass with the updated dependencies, except one for kerberos auth, which fails to properly set up auth locally. 

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,12 +85,20 @@
         <joda.version>2.9.6</joda.version>
         <kafka.connect.storage.common.version>11.1.0-SNAPSHOT</kafka.connect.storage.common.version>
         <maven.release.plugin.version>2.5.3</maven.release.plugin.version>
-        <netty.version>4.1.68.Final</netty.version>
+        <netty.version>4.1.71.Final</netty.version>
         <parquet.version>1.11.1</parquet.version>
         <json.smart.version>2.4.7</json.smart.version>
         <thrift.version>0.13.0</thrift.version>
         <tomcat.version>8.5.65</tomcat.version>
         <protobuf.version>3.19.4</protobuf.version>
+        <!--
+        Note: The following three properties are used to change the version of the
+        respective dependencies within hive. They are changed from the ones inherited from
+        hive for fixing CVEs.
+        -->
+        <groovy.override.version>2.4.21</groovy.override.version>
+        <derby.override.version>10.14.1.0</derby.override.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -85,12 +85,20 @@
         <joda.version>2.9.6</joda.version>
         <kafka.connect.storage.common.version>11.1.0-SNAPSHOT</kafka.connect.storage.common.version>
         <maven.release.plugin.version>2.5.3</maven.release.plugin.version>
-        <netty.version>4.1.68.Final</netty.version>
+        <netty.version>4.1.71.Final</netty.version>
```
kpatelatwork, line 88: @snehashisp I do remember having to upgrade tcnative when I had updated netty as they go hand in hand, could you please check if that also needs an upgrade?

snehashisp: The netty library brings in the required version of tcnative. In this case it is not being shadowed by another version being brought by some other dependency. Also, don't see there being a property for tcnative only in the pom so I think we are good here.

kpatelatwork (approved):
LGTM @snehashisp but please target 11.x branch

snehashisp:
> LGTM @snehashisp but please target 11.x branch

The master snapshot version is to 11.1.0-SNAPSHOT. We need to release these updates as 11.1.0 so master is the correct branch as the 11.1.x branch is not created. I will be merging this to master and then cutting the 11.1.x from there.

PR #255, opened 2022-07-21
Add a dependency on slf4j-reload4j to used child plugins

sidd1809:
## Problem

`slf4j-log4j12` has vulnerabilites and is being replaced with `slf4j-reload4j`

https://confluentinc.atlassian.net/browse/CCMSG-1972

## Solution
Adding a dependency on `slf4j-reload4j` in this "parent" pom, to use in child connector plugins.

- [ ] yes
- [X] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan
Will be used in child connect plugins

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,6 +90,7 @@
         <json.smart.version>2.4.7</json.smart.version>
         <thrift.version>0.13.0</thrift.version>
         <tomcat.version>8.5.65</tomcat.version>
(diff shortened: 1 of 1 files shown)


akash-borkar (approved):
LGTM

PR #280, opened 2022-11-03
bumping hadoop version

pkuntalteam1:
## Problem
[CCMSG-2287](https://confluentinc.atlassian.net/browse/CCMSG-2287)  - Upgrade jettison at kafka-connect-storage-cloud
[CCMSG-2293](https://confluentinc.atlassian.net/browse/CCMSG-2293) - Upgrade org.apache.hadoop:hadoop-common at kafka-connect-storage-cloud

## Solution
bumping hadoop version to 3.2.3 which in turns will bump the jettison version to 1.5.1

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,7 +73,7 @@
         <confluent.log4j.version>1.2.17-cp2</confluent.log4j.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
         <confluent.version>5.4.11-SNAPSHOT</confluent.version>
-        <hadoop.version>2.10.2</hadoop.version>
+        <hadoop.version>3.2.3</hadoop.version>
         <hive.version>1.2.2</hive.version>
         <joda.version>2.9.6</joda.version>
         <licenses.version>5.4.11-SNAPSHOT</licenses.version>


snehashisp (commented):
I dont think we should update hadoop common to from 2 to 3 here. IIUC this repo is used in hdfs2 and hadoop common 3 does not guarantee full compatiblity with hdfs 2.  https://stackoverflow.com/questions/57581636/can-hadoop-3-2-hdfs-client-be-used-to-work-with-hadoop-2-x-hdfs-nodes

pbadani:
> I dont think we should update hadoop common to from 2 to 3 here. IIUC this repo is used in hdfs2 and hadoop common 3 does not guarantee full compatiblity with hdfs 2. https://stackoverflow.com/questions/57581636/can-hadoop-3-2-hdfs-client-be-used-to-work-with-hadoop-2-x-hdfs-nodes

@snehashisp Given the number of CVE's we have, my preference is to bump it up. We can look into not bumping up the common dependency for HDFS2 if that is a concern.

pkuntalteam1:
> I dont think we should update hadoop common to from 2 to 3 here. IIUC this repo is used in hdfs2 and hadoop common 3 does not guarantee full compatiblity with hdfs 2. https://stackoverflow.com/questions/57581636/can-hadoop-3-2-hdfs-client-be-used-to-work-with-hadoop-2-x-hdfs-nodes

this is the specific version which resolves noted vulnerabilities , we will have to check compatibility with hdfs2.

snehashisp:
> @snehashisp Given the number of CVE's we have, my preference is to bump it up. We can look into not bumping up the common dependency for HDFS2 if that is a concern.

In that case we should override this property in whatever connector repo that does not require hdfs 2 compatibility by default, and see if nothing else breaks. I am against doing it here as this is a common shared repo for multiple connectors.

pkuntalteam1:
> > @snehashisp Given the number of CVE's we have, my preference is to bump it up. We can look into not bumping up the common dependency for HDFS2 if that is a concern.
> 
> In that case we should override this property in whatever connector repo that does not require hdfs 2 compatibility by default, and see if nothing else breaks. I am against doing it here as this is a common shared repo for multiple connectors.

Yes .. right .. this is a messy patch. I will workout different/focussed sol for this.

PR #299, opened 2023-01-31
CCLOG-2401: Parse logical types correctly for hive schemas. 

snehashisp:
## Problem
Some logical connect data types are not correctly inferred to their hive counterpart. These include Date and Timestamp datatypes which are inferred as `int` and `long` previously. However in the `orc` writer these are inserted as Timestamp and Dates respectively which causes data with such datatypes to fail when queried in hive. Time logical type can also either be `int` or `long`. 

There is also a bug in the recursive parsing logic where if logical types are not handled if nested inside a struct, array or map. This is because in case of struct the `convertPrimitiveMaybeLogical` is never reached as the recursive calls always goes to the not logical convert method. 

## Solution
Assign the correct hive data types for the logical connect types. The hive logical conversion is not used currently in any of the downstream hdfs connectors which use this package. The current orc writer does not correctly infer logical types hence there should not be any backwards compatibility issues. Regardless this will be released as a minor version bump. 

The recursive logic is fixed and some tests are added. 

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
+++ b/hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
@@ -19,6 +19,7 @@ import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.serde2.typeinfo.DecimalTypeInfo;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfo;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory;
+import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Decimal;
 import org.apache.kafka.connect.data.Field;
 import org.apache.kafka.connect.data.Schema;
@@ -28,6 +29,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import org.apache.kafka.connect.data.Timestamp;
 import org.apache.kafka.connect.errors.ConnectException;
 
 public class HiveSchemaConverter {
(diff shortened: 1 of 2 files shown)


hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
```diff
@@ -77,51 +79,58 @@ public static List<FieldSchema> convertSchemaMaybeLogical(Schema schema) {
...
+        return convertStruct(schema, convertLogical);
       case ARRAY:
-        return convertArray(schema);
+        return convertArray(schema, convertLogical);
       case MAP:
-        return convertMap(schema);
+        return convertMap(schema, convertLogical);
       default:
-        return convertPrimitive(schema);
+        return convertLogical ? convertPrimitiveMaybeLogical(schema) : convertPrimitive(schema);
```
sudeshwasnik, line 92: hi, thanks for the PR. 
QQ - is there a reason we can't just call `convertPrimitiveMaybeLogical(schema)` with `default` in `convert(Schema schem)` ? 
IIUC, unless schema.name() matches any of `Decimal`, `TypeInfoFactory.dateTypeInfo` OR `TypeInfoFactory.timestampTypeInfo ` -> convertPrimitiveMaybeLogical(schema) will call convertPrimitive(schema) internally - so should be the same. 
ie why do we need `convertSchemaMaybeLogical` and `convertSchema` both ?

sudeshwasnik: the above question is not really related to the change though. 

The change looks good, optionally a cleaner solution could be to create a class that implements these `convert` functions. And if initialized with logcial=true, then just overrides default behvaiour to call convertPrimitiveMaybeLogical instead of convertPrimitive. We don't really need to pass the variable to each step in the recursion  (given that convertMap, convertStruct etc doesn't really use it themselves)

sudeshwasnik:
QQ - is this a `major` release? Since it is backward-incompatible fix ? (schema is going to change).

hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
snehashisp: We do need both, right now. Logical types are internally represented with primitives. For example Time, Date, Timestamps are `int`, Decimal is `byte[]`. `convertPrimitive` would result in representation of these as the primitives in hive (int, byte[]) and this actually fails for Orc as while writing we write the logical type. Moving to `convertSchemaMaybeLogical` for everything is something we could do but most places use the primitive approach and it may require additional changes downstream to switch. The changes to Orc is being done in a separate PR in hdfs.

Creating another class makes for better code but will require new instantiations for every schema conversion. That induces requirements on how the API, for example if the schema generated is not being cached or changing frequently it might create a bunch of single use classes. It is better to keep things stateless.

snehashisp:
> QQ - is this a `major` release? Since it is backward-incompatible fix ? (schema is going to change).

Most places downstream, that I know of, use the primitive approach to conversion and hence will not be affected. Orc will move to the new schemas. The existing issues with logical types means most likely they are not being used, hence compatibility should be fine. Thinking of putting this under a minor version bump.

hive/src/main/java/io/confluent/connect/storage/hive/HiveSchemaConverter.java
sudeshwasnik: gotcha! 
thanks!

sudeshwasnik:
@snehashisp gotcha. imo if schema can change with upgrade, we should term it backward-incompatible. If it is, ideally we should make a major release. 
@confluentinc/connect-team1 wdyt ? I don't have a strong opinion, but lets get one additional +1 for minor release.

sudeshwasnik:
The changes LGTM! Nice work @snehashisp !!

pbadani:
> @snehashisp gotcha. imo if schema can change with upgrade, we should term it backward-incompatible. If it is, ideally we should make a major release. @confluentinc/connect-team1 wdyt ? I don't have a strong opinion, but lets get one additional +1 for minor release.

ok with minor release here. more important question is whether it will be a patch or a new version for the connector.

sudeshwasnik (approved):

PR #311, opened 2023-03-08
update jetty version to match kafka-rest, pin additional jetty libraries

janjwerner-confluent:
## Problem
Parent pom defines outdated version of jetty that could lead to bringing outdated, vulnerable dependencies by connectors using kafka-connect-storage pom

## Solution
Bump version of the dependency, specify dependencies in the dependencyManagement to instruct artifacts built using this pom to use the updated versions.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


janjwerner-confluent:
stale, unreviewed since March

PR #314, opened 2023-04-24
CC-19464 | Return the reason for schema incompatibility along with the boolean result.

pbadani:
## Problem
Return the schema incompatibility reason so that it can be logged in downstream connectors to enhance logging.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


aniketshrimal (commented):
Thanks for the PR @pbadani! Two high level comments - 

1. Lets capture each of the schema compatibility scenario as Unit tests, if possible. 
2. We should target this to master branch and cut a new minor version for this change.

aniketshrimal (approved):
LGTM! Thanks @pbadani

PR #347, opened 2024-02-20
Use record.newRecord instead of Creating new SinkRecord

naveenmall11:
## Problem
https://confluent.slack.com/archives/C06HJ2VFQ5T
While projecting we were changing instance of record from InternalSinkRecord to SinkRecord and while reporting we were using modified records which was causing the original schema of messages to be changed while serialising again.

## Solution

Instead of creating a new object, we can use record.newRecord. Functionality wise it will same for connector and while reporting we will be able to report instance of `InternalSinkRecord'

- [ ] yes
- [ ] no

## Test Strategy
Tested the change with reproduction model https://github.com/confluentinc/kafka-docker-playground-internal/blob/5fbb61804e4a282ccd29cc77719bfc8cc14d19ed/connect-connect-aws-s3-sink/s3-sink-repro-194748-unexpected-register-of-schema.sh

After the fix, no unwanted version of schema register was found. 
image
image

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/core/src/main/java/io/confluent/connect/storage/schema/StorageSchemaCompatibility.java
+++ b/core/src/main/java/io/confluent/connect/storage/schema/StorageSchemaCompatibility.java
@@ -371,19 +371,18 @@ public enum StorageSchemaCompatibility implements SchemaCompatibility {
     );
 
     // Just reference comparison here.
-    return projected.getKey() == record.key() && projected.getValue() == record.value()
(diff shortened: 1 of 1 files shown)


b-goyal (approved):
LGTM

PR #367, opened 2024-05-14
fix jackson and snappy CVE

sambhav-jain-16:
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <groupId>io.confluent</groupId>
         <artifactId>common</artifactId>
-        <version>6.1.12</version>
+        <version>6.1.15</version>
     </parent>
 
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -21,7 +21,7 @@
     <parent>
         <groupId>io.confluent</groupId>
         <artifactId>common</artifactId>
-        <version>6.1.12</version>
+        <version>6.1.16-14</version>
```
mukkachaitanya, line 24: I don't think we should be using nano version artifacts for connector releases. They are typically meant for cloud-based projects.

sagarrao12 (approved):
LGTM

PR #370, opened 2024-05-28
CC-27029 Bump joda-time version

subhashiyer9:
## Problem
joda-time version 2.9.6 does not handle the daylight savings for some timezones and was formatting the time incorrectly. 
Example, for America/Santiago timezone, the timezone was formatted incorrectly when the day light savings ended

RCCA-19503

## Solution
Fix the dependency to v2.10.1 which has the fix for same.

- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


sp-gupta (approved):
LGTM

PR #425, opened 2025-04-11
cc-32375-33204 | Upgraded the org.apache.parquet version to 1.15.1

SaiPriyatam6:
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/format/pom.xml
+++ b/format/pom.xml
@@ -31,6 +31,7 @@
         <jackson.groupId>com.fasterxml.jackson.core</jackson.groupId>
         <jackson.package>com.fasterxml.jackson</jackson.package>
         <shade.prefix>shaded.parquet</shade.prefix>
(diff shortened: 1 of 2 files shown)


format/pom.xml
```diff
@@ -31,6 +31,7 @@
         <jackson.groupId>com.fasterxml.jackson.core</jackson.groupId>
         <jackson.package>com.fasterxml.jackson</jackson.package>
         <shade.prefix>shaded.parquet</shade.prefix>
+        <parquet.version>1.15.1</parquet.version>
```
arihant-confluent, line 34: why it is required, shouldn't it be coming from the parent class?

PR #466, opened 2026-03-04
update parent pom

janjwerner-confluent:
## Problem
An outdated pom parent is being used

## Solution
Update pom parent to avoid outdated dependency definitions

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


ahmedashfaq027 (approved):
LGTM.

Quick question: How do we ensure the version is latest always?

PR #486, opened 2026-07-24
feat(INIT-5113): add backup envelope infrastructure with shared config validation

sribaabu:
## Problem
     
  Object storage connectors (S3, GCS, Azure) need shared infrastructure for backup and restore of full Kafka records. The envelope wrapping, schema backup, schema manifest, and config validation logic must be common across all connector
  backends.
  
  ## Solution
  
  Add backup envelope infrastructure to storage-common, shared by all sink and source connectors:
  
  - BackupEnvelope: envelope schema constants and field definitions
  - EnvelopeSchemaBuilder: builds KafkaRecordEnvelope Connect struct (key, value, headers, metadata, schema info)
  - EnvelopeTransformer: wraps SinkRecords in envelope structs during backup
  - BackupWrapperExtractor: unwraps BackupWrapper metadata from converter output
  - SchemaBackupOrchestrator: triggers schema file backup with 3-level dedup and reference tree traversal
  - ObjectStoreSchemaBackupStore: writes schema files (.avsc/.proto/.json) and entry files (.entry.json)
  - SchemaManifest: schema entry and reference entry data model
  - BackupModeValidator: shared 3-tier config validation (fail/warn/info) for all backends
  - ConverterTypeDetector: maps converter class names to schema type tags
  - StorageSinkConnectorConfig: adds mode=BACKUP_FULL_RECORD config with ConfigDef validation
  
  - [x] yes
  - [ ] no
  
  ##### If yes, where? 
  Storage-cloud (S3 sink connector, PR pending), object-store-source (source connector, PR pending), GCS and Azure sink connectors (future PRs).
  
  ## Test Strategy
  
  ##### Testing done:
  - [x] Unit tests
  - [ ] Integration tests
  - [ ] System tests
  - [x] Manual tests

  154 unit tests (120 core + 34 format), 0 regressions. Covers envelope building, schema backup dedup, reference parsing, converter type detection, envelope transformer, schema backup orchestrator, and wrapper extraction (17 scenarios). E2E
  validated via Docker.
  
  ## Release Plan
  
  Merging to master via INIT-5113-phase2. Must merge before storage-cloud and object-store-source PRs. Part of Backup & Restore feature (INIT-5113).

### DIFF
--- a/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
+++ b/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
@@ -100,6 +100,14 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
   public static final int SCHEMA_CACHE_SIZE_DEFAULT = 1000;
   public static final String SCHEMA_CACHE_SIZE_DISPLAY = "Schema Cache Size";
 
+  public static final String FORMAT_JSON_SCHEMA_ENABLE_CONFIG = "format.json.schema.enable";
+  public static final boolean FORMAT_JSON_SCHEMA_ENABLE_DEFAULT = false;
+  private static final String FORMAT_JSON_SCHEMA_ENABLE_DOC =
+      "Whether to embed the Connect schema in JSON output files. When true, each line "
+      + "is written as {\"schema\":...,\"payload\":...}. Required for BACKUP_FULL_RECORD "
+      + "mode with JsonFormat to enable full restore round-trip.";
+  private static final String FORMAT_JSON_SCHEMA_ENABLE_DISPLAY = "Enable Embedded JSON Schema";
+
   public static final String ENHANCED_AVRO_SCHEMA_SUPPORT_CONFIG = "enhanced.avro.schema.support";
   public static final boolean ENHANCED_AVRO_SCHEMA_SUPPORT_DEFAULT = true;
   public static final String ENHANCED_AVRO_SCHEMA_SUPPORT_DOC =
@@ -144,6 +152,35 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
       + "supported configurations are NONE, BACKWARD, FORWARD and FULL.";
   public static final String SCHEMA_COMPATIBILITY_DEFAULT = "NONE";
   public static final String SCHEMA_COMPATIBILITY_DISPLAY = "Schema Compatibility";
+
+  // Mode group
+  public static final String MODE_CONFIG = "mode";
+  public static final String MODE_DEFAULT = Mode.GENERIC.name();
+  public static final String MODE_DOC =
+      "The connector's operation mode. "
+      + "GENERIC: standard sink behavior. "
+      + "BACKUP_FULL_RECORD: consolidates key, value, headers and metadata "
+      + "into a single envelope record per message with pristine schema preservation.";
+
+  /**
+   * Connector operation mode. Determines whether the sink operates in
+   * standard mode or backup envelope mode.
+   */
+  public enum Mode {
+    GENERIC,
+    BACKUP_FULL_RECORD;
+
+    public static Mode of(String name) {
+      for (Mode m : values()) {
+        if (m.name().equalsIgnoreCase(name)) {
+          return m;
+        }
+      }
+      throw new org.apache.kafka.common.config.ConfigException(
+          MODE_CONFIG, name,
+          "Invalid mode. Valid values: " + java.util.Arrays.toString(values()));
+    }
+  }
 
   // CHECKSTYLE:OFF
   public static final ConfigDef.Recommender schemaCompatibilityRecommender =
@@ -289,40 +326,45 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
           FILENAME_OFFSET_ZERO_PAD_WIDTH_DISPLAY
       );
 
-      configDef.define(
-          AVRO_CODEC_CONFIG,
-          Type.STRING,
-          AVRO_CODEC_DEFAULT,
-          ConfigDef.ValidString.in(AVRO_SUPPORTED_CODECS),
-          Importance.LOW,
-          AVRO_CODEC_DOC,
-          group,
-          ++orderInGroup,
-          Width.MEDIUM,
-          AVRO_CODEC_DISPLAY,
-          avroRecommender
-      );
-
-      configDef.define(
-          ALLOW_OPTIONAL_MAP_KEYS,
-          Type.BOOLEAN,
-          ALLOW_OPTIONAL_MAP_KEYS_DEFAULT,
-          Importance.LOW,
-          ALLOW_OPTIONAL_MAP_KEYS_DOC,
-          group,
-          ++orderInGroup,
-          Width.SHORT,
-          ALLOW_OPTIONAL_MAP_KEYS_DISPLAY
-      );
-
+      addCodecAndMapConfigs(configDef, avroRecommender, group, orderInGroup);
     }
+    addSchemaAndModeConfigs(configDef);
+    return configDef;
+  }
 
+  private static void addCodecAndMapConfigs(
+      ConfigDef configDef, ConfigDef.Recommender avroRecommender,
+      String group, int orderInGroup) {
+    configDef.define(
+        AVRO_CODEC_CONFIG,
+        Type.STRING,
+        AVRO_CODEC_DEFAULT,
+        ConfigDef.ValidString.in(AVRO_SUPPORTED_CODECS),
+        Importance.LOW,
+        AVRO_CODEC_DOC,
+        group,
+        ++orderInGroup,
+        Width.MEDIUM,
+        AVRO_CODEC_DISPLAY,
+        avroRecommender
+    );
+    configDef.define(
+        ALLOW_OPTIONAL_MAP_KEYS,
+        Type.BOOLEAN,
+        ALLOW_OPTIONAL_MAP_KEYS_DEFAULT,
+        Importance.LOW,
+        ALLOW_OPTIONAL_MAP_KEYS_DOC,
+        group,
+        ++orderInGroup,
+        Width.SHORT,
+        ALLOW_OPTIONAL_MAP_KEYS_DISPLAY
+    );
+  }
+
+  private static void addSchemaAndModeConfigs(ConfigDef configDef) {
     {
-      // Define Schema configuration group
       final String group = "Schema";
       int orderInGroup = 0;
-
-      // Define Schema configuration group
       configDef.define(
           SCHEMA_COMPATIBILITY_CONFIG,
           Type.STRING,
@@ -336,7 +378,22 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
           schemaCompatibilityRecommender
       );
     }
-    return configDef;
+    {
+      final String group = "Mode";
+      int orderInGroup = 0;
+      configDef.define(
+          MODE_CONFIG,
+          Type.STRING,
+          MODE_DEFAULT,
+          ConfigDef.ValidString.in(Mode.GENERIC.name(), Mode.BACKUP_FULL_RECORD.name()),
+          Importance.MEDIUM,
+          MODE_DOC,
+          group,
+          ++orderInGroup,
+          Width.SHORT,
+          "Mode"
+      );
+    }
   }
 
   /**
@@ -368,6 +425,17 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
         Width.MEDIUM,
         PARQUET_CODEC_DISPLAY,
         parquetRecommender
+    );
+    configDef.define(
+        FORMAT_JSON_SCHEMA_ENABLE_CONFIG,
+        Type.BOOLEAN,
+        FORMAT_JSON_SCHEMA_ENABLE_DEFAULT,
+        Importance.LOW,
+        FORMAT_JSON_SCHEMA_ENABLE_DOC,
+        group,
+        ++orderInGroup,
+        Width.SHORT,
+        FORMAT_JSON_SCHEMA_ENABLE_DISPLAY
     );
   }
 
@@ -432,5 +500,24 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
     props.put(CONNECT_META_DATA_CONFIG, get(CONNECT_META_DATA_CONFIG));
     props.put(ALLOW_OPTIONAL_MAP_KEYS, get(ALLOW_OPTIONAL_MAP_KEYS));
     return new AvroDataConfig(props);
+  }
+
+  public Mode mode() {
+    return Mode.of(getString(MODE_CONFIG));
+  }
+
+  public boolean isBackupMode() {
+    return mode() == Mode.BACKUP_FULL_RECORD;
+  }
+
+  public boolean isJsonSchemaEmbedded() {
+    return getBoolean(FORMAT_JSON_SCHEMA_ENABLE_CONFIG);
+  }
+
+  public String getEffectiveSchemaCompatibility() {
+    if (isBackupMode()) {
+      return "NONE";
+    }
+    return getString(SCHEMA_COMPATIBILITY_CONFIG);
   }
 }
--- a/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
+++ b/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
@@ -378,23 +378,25 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
           schemaCompatibilityRecommender
       );
     }
-    {
-      final String group = "Mode";
-      int orderInGroup = 0;
-      configDef.define(
-          MODE_CONFIG,
-          Type.STRING,
-          MODE_DEFAULT,
-          ConfigDef.ValidString.in(Mode.GENERIC.name(), Mode.BACKUP_FULL_RECORD.name()),
-          Importance.MEDIUM,
-          MODE_DOC,
-          group,
-          ++orderInGroup,
-          Width.SHORT,
-          "Mode"
-      );
-    }
+    addModeGroup(configDef);
     return configDef;
+  }
+
+  private static void addModeGroup(ConfigDef configDef) {
+    final String group = "Mode";
+    int orderInGroup = 0;
+    configDef.define(
+        MODE_CONFIG,
+        Type.STRING,
+        MODE_DEFAULT,
+        ConfigDef.ValidString.in(Mode.GENERIC.name(), Mode.BACKUP_FULL_RECORD.name()),
+        Importance.MEDIUM,
+        MODE_DOC,
+        group,
+        ++orderInGroup,
+        Width.SHORT,
+        "Mode"
+    );
   }
 
   /**
--- a/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
+++ b/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
@@ -393,6 +393,17 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
           Width.SHORT,
           "Mode"
       );
+      configDef.define(
+          FORMAT_JSON_SCHEMA_ENABLE_CONFIG,
+          Type.BOOLEAN,
+          FORMAT_JSON_SCHEMA_ENABLE_DEFAULT,
+          Importance.LOW,
+          FORMAT_JSON_SCHEMA_ENABLE_DOC,
+          group,
+          ++orderInGroup,
+          Width.SHORT,
+          FORMAT_JSON_SCHEMA_ENABLE_DISPLAY
+      );
     }
   }
 
@@ -425,17 +436,6 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
         Width.MEDIUM,
         PARQUET_CODEC_DISPLAY,
         parquetRecommender
-    );
-    configDef.define(
-        FORMAT_JSON_SCHEMA_ENABLE_CONFIG,
-        Type.BOOLEAN,
-        FORMAT_JSON_SCHEMA_ENABLE_DEFAULT,
-        Importance.LOW,
-        FORMAT_JSON_SCHEMA_ENABLE_DOC,
-        group,
-        ++orderInGroup,
-        Width.SHORT,
-        FORMAT_JSON_SCHEMA_ENABLE_DISPLAY
     );
   }
 
--- a/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
+++ b/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
@@ -183,6 +183,21 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
       new SchemaCompatibilityRecommender();
   // CHECKSTYLE:ON
 
+  private static final ConfigDef.Recommender MODE_RECOMMENDER = new ConfigDef.Recommender() {
+    private final List<Object> values =
+        Arrays.asList(Mode.GENERIC.name(), Mode.BACKUP_FULL_RECORD.name());
+
+    @Override
+    public List<Object> validValues(String name, Map<String, Object> parsedConfig) {
+      return values;
+    }
+
+    @Override
+    public boolean visible(String name, Map<String, Object> parsedConfig) {
+      return true;
+    }
+  };
+
   /**
    * Create a new configuration definition.
    *
@@ -352,7 +367,7 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
           FORMAT_JSON_SCHEMA_ENABLE_CONFIG,
           Type.BOOLEAN,
           FORMAT_JSON_SCHEMA_ENABLE_DEFAULT,
-          Importance.LOW,
+          Importance.MEDIUM,
           FORMAT_JSON_SCHEMA_ENABLE_DOC,
           group,
           ++orderInGroup,
@@ -395,7 +410,8 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
         group,
         ++orderInGroup,
         Width.SHORT,
-        "Mode"
+        "Mode",
+        MODE_RECOMMENDER
     );
   }
 
--- a/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
+++ b/core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
@@ -162,10 +162,6 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
       + "BACKUP_FULL_RECORD: consolidates key, value, headers and metadata "
       + "into a single envelope record per message with pristine schema preservation.";
 
-  /**
-   * Connector operation mode. Determines whether the sink operates in
-   * standard mode or backup envelope mode.
-   */
   public enum Mode {
     GENERIC,
     BACKUP_FULL_RECORD;
@@ -326,45 +322,49 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
           FILENAME_OFFSET_ZERO_PAD_WIDTH_DISPLAY
       );
 
-      addCodecAndMapConfigs(configDef, avroRecommender, group, orderInGroup);
+      configDef.define(
+          AVRO_CODEC_CONFIG,
+          Type.STRING,
+          AVRO_CODEC_DEFAULT,
+          ConfigDef.ValidString.in(AVRO_SUPPORTED_CODECS),
+          Importance.LOW,
+          AVRO_CODEC_DOC,
+          group,
+          ++orderInGroup,
+          Width.MEDIUM,
+          AVRO_CODEC_DISPLAY,
+          avroRecommender
+      );
+
+      configDef.define(
+          ALLOW_OPTIONAL_MAP_KEYS,
+          Type.BOOLEAN,
+          ALLOW_OPTIONAL_MAP_KEYS_DEFAULT,
+          Importance.LOW,
+          ALLOW_OPTIONAL_MAP_KEYS_DOC,
+          group,
+          ++orderInGroup,
+          Width.SHORT,
+          ALLOW_OPTIONAL_MAP_KEYS_DISPLAY
+      );
+
+      configDef.define(
+          FORMAT_JSON_SCHEMA_ENABLE_CONFIG,
+          Type.BOOLEAN,
+          FORMAT_JSON_SCHEMA_ENABLE_DEFAULT,
+          Importance.LOW,
+          FORMAT_JSON_SCHEMA_ENABLE_DOC,
+          group,
+          ++orderInGroup,
+          Width.SHORT,
+          FORMAT_JSON_SCHEMA_ENABLE_DISPLAY
+      );
     }
-    addSchemaAndModeConfigs(configDef);
-    return configDef;
-  }
-
-  private static void addCodecAndMapConfigs(
-      ConfigDef configDef, ConfigDef.Recommender avroRecommender,
-      String group, int orderInGroup) {
-    configDef.define(
-        AVRO_CODEC_CONFIG,
-        Type.STRING,
-        AVRO_CODEC_DEFAULT,
-        ConfigDef.ValidString.in(AVRO_SUPPORTED_CODECS),
-        Importance.LOW,
-        AVRO_CODEC_DOC,
-        group,
-        ++orderInGroup,
-        Width.MEDIUM,
-        AVRO_CODEC_DISPLAY,
-        avroRecommender
-    );
-    configDef.define(
-        ALLOW_OPTIONAL_MAP_KEYS,
-        Type.BOOLEAN,
-        ALLOW_OPTIONAL_MAP_KEYS_DEFAULT,
-        Importance.LOW,
-        ALLOW_OPTIONAL_MAP_KEYS_DOC,
-        group,
-        ++orderInGroup,
-        Width.SHORT,
-        ALLOW_OPTIONAL_MAP_KEYS_DISPLAY
-    );
-  }
-
-  private static void addSchemaAndModeConfigs(ConfigDef configDef) {
     {
       final String group = "Schema";
       int orderInGroup = 0;
+
+      // Define Schema configuration group
       configDef.define(
           SCHEMA_COMPATIBILITY_CONFIG,
           Type.STRING,
@@ -393,18 +393,8 @@ public class StorageSinkConnectorConfig extends AbstractConfig implements Compos
           Width.SHORT,
           "Mode"
       );
-      configDef.define(
-          FORMAT_JSON_SCHEMA_ENABLE_CONFIG,
-          Type.BOOLEAN,
-          FORMAT_JSON_SCHEMA_ENABLE_DEFAULT,
-          Importance.LOW,
-          FORMAT_JSON_SCHEMA_ENABLE_DOC,
-          group,
-          ++orderInGroup,
-          Width.SHORT,
-          FORMAT_JSON_SCHEMA_ENABLE_DISPLAY
-      );
     }
+    return configDef;
   }
 
   /**
--- a/core/src/main/java/io/confluent/connect/storage/StorageSinkTestBase.java
+++ b/core/src/main/java/io/confluent/connect/storage/StorageSinkTestBase.java
@@ -21,6 +21,7 @@ import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.common.metrics.PluginMetrics;
 import org.apache.kafka.connect.sink.ErrantRecordReporter;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.sink.SinkTaskContext;
@@ -253,6 +254,11 @@ public class StorageSinkTestBase {
     @Override
     public ErrantRecordReporter errantRecordReporter() {
       throw new UnsupportedOperationException("ErrantRecordReporter is undefined for this class");
+    }
+
+    @Override
+    public PluginMetrics pluginMetrics() {
+      return null;
     }
   }
 
--- a/core/src/main/java/io/confluent/connect/storage/StorageSinkTestBase.java
+++ b/core/src/main/java/io/confluent/connect/storage/StorageSinkTestBase.java
@@ -21,7 +21,6 @@ import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Struct;
-import org.apache.kafka.common.metrics.PluginMetrics;
 import org.apache.kafka.connect.sink.ErrantRecordReporter;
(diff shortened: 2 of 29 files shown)


core/src/main/java/io/confluent/connect/storage/backup/BackupEnvelope.java
```diff
@@ -0,0 +1,129 @@
...
+
+  // Envelope field names
+  public static final String FIELD_KEY = "key";
+  public static final String FIELD_VALUE = "value";
+  public static final String FIELD_HEADERS = "headers";
+  public static final String FIELD_TOPIC = "topic";
+  public static final String FIELD_PARTITION = "partition";
+  public static final String FIELD_OFFSET = "offset";
+  public static final String FIELD_TIMESTAMP = "timestamp";
+  public static final String FIELD_TIMESTAMP_TYPE = "timestampType";
```
subhashiyer9, line 42: What are the different values for `timestampType`?

core/src/main/java/io/confluent/connect/storage/backup/BackupEnvelope.java
```diff
@@ -0,0 +1,129 @@
...
+   */
+  public static boolean isSrBackedType(String schemaType) {
+    return SchemaBackupConfig.isSrBackedType(schemaType);
+  }
+
+  /**
+   * Returns the file extension for a schema type.
+   */
+  public static String extensionForType(String schemaType) {
+    if (TYPE_AVRO.equals(schemaType)) {
```
subhashiyer9, line 115: Do these checks need to be made case insensitive?

core/src/main/java/io/confluent/connect/storage/backup/BackupEnvelope.java
```diff
@@ -0,0 +1,129 @@
...
+      return EXT_AVRO;
+    }
+    if (TYPE_PROTOBUF.equals(schemaType)) {
+      return EXT_PROTOBUF;
+    }
+    if (TYPE_JSON_SCHEMA.equals(schemaType) || TYPE_JSON.equals(schemaType)) {
+      return EXT_JSON;
+    }
+    return EXT_DEFAULT;
```
subhashiyer9, line 124: What is the format used in the default `EXT_DEFAULT`?

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
```diff
@@ -0,0 +1,293 @@
...
+          + "BACKUP_FULL_RECORD mode. ByteArrayFormat does not support "
+          + "structured schema metadata required for envelope wrapping. "
+          + "Use AvroFormat, JsonFormat, or ParquetFormat instead.");
+    }
+  }
+
+  private static void validateJsonFormatSchemaEnable(
+      String formatClassName, boolean jsonSchemaEmbedded,
+      List<String> errors) {
+    if (FORMAT_SIMPLE_NAME_JSON.equals(formatClassName) && !jsonSchemaEmbedded) {
```
subhashiyer9, line 138: What is the alternative for customers running without a schema?

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
```diff
@@ -0,0 +1,293 @@
...
+    if (!BackupEnvelope.isSrBackedType(schemaType)) {
+      return;
+    }
+    String configKey = converterPrefix + "."
+        + BackupEnvelope.SCHEMA_BACKUP_ENABLED_CONFIG;
+    if (!"true".equalsIgnoreCase(configs.get(configKey))) {
+      errors.add(converterPrefix + " uses SR-backed converter ("
+          + converterClass + ") but " + configKey + " is not set to true. "
+          + "Without this config, backup will NOT preserve schema metadata "
+          + "and restore will produce corrupted data. "
```
subhashiyer9, line 161: > restore will produce corrupted data.

Just trying to understand the nature of corruption here

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
subhashiyer9: Possibly, it can fall back to backing up just the value without any schema.

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
```diff
@@ -0,0 +1,293 @@
...
+          + "JsonSchemaConverter with AvroFormat.");
+    }
+  }
+
+  private static void warnParquetCompression(
+      Map<String, String> configs, String formatClassName) {
+    if (!FORMAT_SIMPLE_NAME_PARQUET.equals(formatClassName)) {
+      return;
+    }
+    String codec = configs.get("parquet.codec");
```
subhashiyer9, line 241: Any reason why this is recommended? What is the error with codec != none?

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
```diff
@@ -0,0 +1,293 @@
...
+    if (codec != null && !"none".equalsIgnoreCase(codec)) {
+      log.warn("parquet.codec={} detected with backup mode. "
+          + "parquet.codec=none is recommended for backup mode. "
+          + "Compressed parquet files may cause restore failures.", codec);
+    }
+  }
+
+  private static void warnHeaderConverter(Map<String, String> configs) {
+    String headerConverter = configs.get("header.converter");
+    if (headerConverter == null
```
subhashiyer9, line 251: `SimpleHeaderConverter` is the default header converter in use.
What is the problem in that case? Why `ByteArrayConverter` is the only recommended one?

core/src/main/java/io/confluent/connect/storage/backup/BackupReferenceParser.java
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * Parse the direct references JSON using globalIds from the reference tree.
+   * Only direct references are returned (not transitive ones).
+   *
+   * @param directRefsJson the JSON from Wrapper's directRefs field
+   * @param referenceTreeJson the JSON from Wrapper's referenceTree field
+   * @return list of direct reference entries, or empty list if either is null
+   */
+  @SuppressWarnings("unchecked")
+  public static List<SchemaManifest.SchemaReferenceEntry> parseDirectRefsToManifestEntries(
```
subhashiyer9, line 51: Could you help with an example of what it parses?

core/src/main/java/io/confluent/connect/storage/backup/ConverterTypeDetector.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  private ConverterTypeDetector() {
+  }
+
+  /**
+   * Register a custom converter type mapping. Call during connector startup
+   * for converters not in the built-in list.
+   */
+  public static void register(String converterClassName, String schemaType) {
```
subhashiyer9, line 62: Is there any known use case of this `register` in the current version?

core/src/main/java/io/confluent/connect/storage/backup/ConverterTypeDetector.java
```diff
@@ -0,0 +1,91 @@
...
+        BackupEnvelope.TYPE_PROTOBUF);
+    KNOWN_TYPES.put("io.confluent.connect.json.JsonSchemaConverter",
+        BackupEnvelope.TYPE_JSON_SCHEMA);
+    KNOWN_TYPES.put("org.apache.kafka.connect.storage.StringConverter",
+        BackupEnvelope.TYPE_STRING);
+    KNOWN_TYPES.put("org.apache.kafka.connect.converters.IntegerConverter",
+        BackupEnvelope.TYPE_INT32);
+    KNOWN_TYPES.put("org.apache.kafka.connect.converters.LongConverter",
+        BackupEnvelope.TYPE_INT64);
+    KNOWN_TYPES.put("org.apache.kafka.connect.converters.ShortConverter",
```
subhashiyer9, line 42: Are all of these converters like LongConverter, ShortConverter etc relevant and in use?

core/src/main/java/io/confluent/connect/storage/backup/ConverterTypeDetector.java
```diff
@@ -0,0 +1,91 @@
...
+   *
+   * @param converterClass fully qualified converter class name
+   * @param config task config map (used for JsonConverter schemas.enable check)
+   * @param converterPrefix converter config prefix (e.g. "key.converter")
+   * @return the schema type tag, or "UNKNOWN" if not recognized
+   */
+  public static String detectSchemaType(
+      String converterClass, Map<String, String> config, String converterPrefix) {
+    if (converterClass == null) {
+      return BackupEnvelope.TYPE_NONE;
```
subhashiyer9, line 77: What does `BackupEnvelope.TYPE_NONE;` mean?

core/src/main/java/io/confluent/connect/storage/backup/EnvelopeSchemaBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+
+  public static final Schema HEADERS_ARRAY_SCHEMA =
+      SchemaBuilder.array(HEADER_SCHEMA).optional().build();
+
+  public static Schema buildEnvelopeSchema(
+      Schema keySchema, Schema valueSchema,
+      String keySchemaType, String valueSchemaType) {
+
+    SchemaBuilder b = SchemaBuilder.struct().name(BackupEnvelope.NAME);
+    b.field(BackupEnvelope.FIELD_KEY,
```
subhashiyer9, line 55: How did you arrive at the default of OPTIONAL_STRING_SCHEMA for keySchema?

core/src/main/java/io/confluent/connect/storage/tools/SchemaSourceConnector.java
```diff
@@ -16,7 +16,7 @@
 package io.confluent.connect.storage.tools;
 
 import org.apache.kafka.common.config.ConfigDef;
-import org.apache.kafka.common.utils.AppInfoParser;
+import org.apache.kafka.common.utils.internals.AppInfoParser;
```
subhashiyer9: What triggered this change?
Does this result in incompatibilities with any CP version?

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
```diff
@@ -289,40 +326,45 @@ public static ConfigDef newConfigDef(
...
-          ALLOW_OPTIONAL_MAP_KEYS_DEFAULT,
-          Importance.LOW,
-          ALLOW_OPTIONAL_MAP_KEYS_DOC,
-          group,
-          ++orderInGroup,
-          Width.SHORT,
-          ALLOW_OPTIONAL_MAP_KEYS_DISPLAY
-      );
-
+      addCodecAndMapConfigs(configDef, avroRecommender, group, orderInGroup);
```
subhashiyer9, line 329: nit: Why did you decide to move these to a new method? codec and map configs looks unrelated to each other
I am suspecting this may be in response to checkstyle errors, in which case you can suppress the warnings for this class.

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
```diff
@@ -369,6 +426,17 @@ public static <T extends ConfigDef.Recommender & ConfigDef.Validator> void enabl
         PARQUET_CODEC_DISPLAY,
         parquetRecommender
     );
+    configDef.define(
+        FORMAT_JSON_SCHEMA_ENABLE_CONFIG,
```
subhashiyer9, line 430: Is this a new config being exposed?

subhashiyer9 (commented):
Added some early high level comments.
I will take a deeper second pass later.

format/src/main/java/io/confluent/connect/storage/format/backup/BackupWrapperExtractor.java
```diff
@@ -0,0 +1,187 @@
...
+    if (schema == null && data == null) {
+      return Unwrapped.tombstone();
+    }
+    if (schema == null) {
+      return unwrapSchemaless(data);
+    }
+    return Unwrapped.passthrough(data, schema, schemaTypeDefault);
+  }
+
+  private static Unwrapped unwrapFromWrapper(Struct w, Schema schema) {
```
subhashiyer9, line 58: nit: Rename `w`

format/src/main/java/io/confluent/connect/storage/format/backup/BackupWrapperExtractor.java
```diff
@@ -0,0 +1,187 @@
...
+    if (BackupWrapper.isWrapper(schema) && data instanceof Struct) {
+      return unwrapFromWrapper((Struct) data, schema);
+    }
+    if (schema == null && data == null) {
+      return Unwrapped.tombstone();
+    }
+    if (schema == null) {
+      return unwrapSchemaless(data);
+    }
+    return Unwrapped.passthrough(data, schema, schemaTypeDefault);
```
subhashiyer9, line 55: QQ: Is there any realistic scenario, where pass through will be called?

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
```diff
@@ -0,0 +1,293 @@
...
+  public static void logSinkStartupSummary(
+      Map<String, String> configs,
+      String formatClassName, String keyType, String valueType) {
+    String valConverter = configs.get(BackupEnvelope.VALUE_CONVERTER_CONFIG);
+    String keyConverter = configs.get(BackupEnvelope.KEY_CONVERTER_CONFIG);
+    String backupEnabled = configs.get(
+        BackupEnvelope.VALUE_CONVERTER_CONFIG + "."
+        + BackupEnvelope.SCHEMA_BACKUP_ENABLED_CONFIG);
+    log.info("Backup mode started: format={}, keyConverter={} (type={}), "
+        + "valueConverter={} (type={}), schema.backup.enabled={}",
```
subhashiyer9, line 118: `schema.backup.enabled` is to be set only on the Value converter, or all converters?
What happens when one of the converters has it set vs rest not?

format/src/main/java/io/confluent/connect/storage/format/backup/EnvelopeTransformer.java
```diff
@@ -0,0 +1,131 @@
...
+   * Wraps a batch of SinkRecords in envelope structs.
+   * Also backs up schema files for any new schemas encountered.
+   *
+   * @param records the records from {@code put()}
+   * @return wrapped records ready for {@code TopicPartitionWriter}
+   */
+  public Collection<SinkRecord> wrapAll(Collection<SinkRecord> records) {
+    List<SinkRecord> wrapped = new ArrayList<>(records.size());
+    for (SinkRecord record : records) {
+      wrapped.add(wrap(record));
```
subhashiyer9, line 75: How are any malformed or invalid records handled?
What happens to tombstone records with specific behaviour.on.null configured?

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
```diff
@@ -0,0 +1,293 @@
...
+      Map<String, String> configs,
+      String formatClassName,
+      boolean jsonSchemaEmbedded) {
+    List<String> errors = new ArrayList<>();
+
+    validateByteArrayFormat(formatClassName, errors);
+    validateJsonFormatSchemaEnable(formatClassName, jsonSchemaEmbedded, errors);
+    validateSchemaBackupEnabled(configs, BackupEnvelope.KEY_CONVERTER_CONFIG, errors);
+    validateSchemaBackupEnabled(configs, BackupEnvelope.VALUE_CONVERTER_CONFIG, errors);
+    validateTransformsRejected(configs, errors);
```
subhashiyer9, line 75: Also, add validators on supported partitioners.

format/src/main/java/io/confluent/connect/storage/format/backup/SchemaBackupOrchestrator.java
```diff
@@ -0,0 +1,178 @@
...
+
+  @SuppressWarnings("unchecked")
+  private void backupSingleReference(
+      String topic, Map<String, Object> node,
+      Map<String, Map<String, Object>> tree, Set<Integer> visited) {
+    int globalId = node.get(BackupEnvelope.REF_FIELD_GLOBAL_ID) instanceof Number
+        ? ((Number) node.get(BackupEnvelope.REF_FIELD_GLOBAL_ID)).intValue() : 0;
+    if (globalId <= 0 || !visited.add(globalId)) {
+      if (globalId <= 0) {
+        log.warn("Skipping reference with invalid globalId={} in topic={}",
```
subhashiyer9, line 129: Can we throw an error here?
I believe this will never occur and this just a defensive check?
IMO we can fail here than silently processing incorrectly

format/src/main/java/io/confluent/connect/storage/format/backup/BackupWrapperExtractor.java
```diff
@@ -0,0 +1,187 @@
...
+  private static Unwrapped unwrapSchemaless(Object data) {
+    String stringData;
+    if (data instanceof Map || data instanceof List) {
+      try {
+        stringData = OBJECT_MAPPER.writeValueAsString(data);
+      } catch (JsonProcessingException e) {
+        throw new DataException("Failed to serialize schemaless data as JSON", e);
+      }
+    } else {
+      stringData = data != null ? data.toString() : null;
```
subhashiyer9, line 91: Is there a similar pattern used for schemaless converters of type other than Map and List?
I would assume SR converters will have an example of this?

core/src/main/java/io/confluent/connect/storage/StorageSinkTestBase.java
```diff
@@ -254,6 +255,11 @@ public void requestCommit() {}
     public ErrantRecordReporter errantRecordReporter() {
       throw new UnsupportedOperationException("ErrantRecordReporter is undefined for this class");
     }
+
+    @Override
+    public PluginMetrics pluginMetrics() {
+      return null;
+    }
```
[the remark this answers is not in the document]
sribaabu: Added as part of the version bump, Reverted.

core/src/main/java/io/confluent/connect/storage/tools/SchemaSourceConnector.java
sribaabu: Yes this is as part of the CP version change, Reverted.

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
sribaabu: Yes, This is required for when JsonFormat is used to dump files, This forces the Json Writers to write with the Schemas embedded per JSON entry, Without that, Envelope based restore is broken and cannot infer the Envelope Structure.

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
```diff
@@ -369,6 +426,17 @@ public static <T extends ConfigDef.Recommender & ConfigDef.Validator> void enabl
...
+        Type.BOOLEAN,
+        FORMAT_JSON_SCHEMA_ENABLE_DEFAULT,
+        Importance.LOW,
+        FORMAT_JSON_SCHEMA_ENABLE_DOC,
+        group,
+        ++orderInGroup,
+        Width.SHORT,
+        FORMAT_JSON_SCHEMA_ENABLE_DISPLAY
+    );
```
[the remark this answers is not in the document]
sribaabu: Resolved.

core/src/main/java/io/confluent/connect/storage/backup/ObjectStoreSchemaBackupStore.java
```diff
@@ -0,0 +1,111 @@
...
+      List<SchemaManifest.SchemaReferenceEntry> references) {
+    if (schemaKey == null || schemaKey.isEmpty()) {
+      log.warn("No schema key for topic={}, skipping backup", topic);
+      return;
+    }
+    String dedupKey = topic + DEDUP_KEY_SEPARATOR + schemaKey;
+
+    if (backedUpKeys.contains(dedupKey)) {
+      return;
+    }
```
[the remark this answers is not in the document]
sribaabu: This is Outdated.

core/src/main/java/io/confluent/connect/storage/backup/BackupEnvelope.java
sribaabu: Current known values from Kafka's org.apache.kafka.common.record.TimestampType enum: NO_TIMESTAMP_TYPE, CREATE_TIME, LOG_APPEND_TIME. Currently it is information only.  Kept as OPTIONAL_STRING_SCHEMA rather than a Connect enum. But let me know what you think.

core/src/main/java/io/confluent/connect/storage/backup/BackupEnvelope.java
sribaabu: Made it case-insensitive as a defensive fix. Callers today feed canonical TYPE_* constants that we wrote. But added to cover any future caller outside our code.

core/src/main/java/io/confluent/connect/storage/backup/BackupEnvelope.java
sribaabu: EXT_DEFAULT was a ".schema" placeholder for fallback, so hitting it is not possible with current SR supported types. Switched to a fail-fast DataException on unsupported types and dropped the EXT_DEFAULT.

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
sribaabu: The flag isn't about the record's value/key schema, it's about the envelope. Backup wraps every record in a KafkaRecordEnvelope Struct. JsonFormat  delegates to JsonConverter with schemas.enable = format.json.schema.enable, so without the flag the file has no schema header and restore cannot rebuild the Struct - key, headers, offset, etc. are all lost.

Schemaless values still work fine: the raw Map goes inside the schemaful envelope. AvroFormat and ParquetFormat don't need this flag since they carry schema natively.

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
sribaabu: schema.backup.enabled=true is what makes the SR converter emit the BackupWrapper with raw schema, schema ID, subject, and references. Without it, the envelope lands with null schema IDs/subjects and no .avsc files on disk, so restore cannot re-register the original schemas and downstream consumers get mismatched IDs or unresolvable schemas

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
sribaabu: Backup and restore doesn't support compression end-to-end (sink writes .snappy.parquet, restore matcher only accepts .parquet, files silently skipped), so parquet.codec must be explicitly set to none. Fails on both. unset (defaults to snappy) and any non-none value and I promoted this warn to validation error now.

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
sribaabu: Yeah, SimpleHeaderConverter is fine for plain UTF-8 text but loses byte-fidelity for binary, JSON with whitespace, high-precision numbers and few other Map/ Struct case etc. ByteArrayConverter is recommended because it does zero  interpretation - raw byte[] passthrough. Full breakdown in our header-handling doc

https://confluentinc.atlassian.net/wiki/spaces/~7120208d0cb1e7ae434cebae47b946ca37cc37/pages/5446926351/Backup+Restore+Header+Handling

core/src/main/java/io/confluent/connect/storage/backup/BackupReferenceParser.java
sribaabu: Say an Avro schema references com.example.Address, which itself references
  com.example.Country. The Wrapper carries two JSON fields from SR:

  directRefsJson (direct references only, no globalIds):
    [{"name":"com.example.Address","subject":"address-value","version":1}]

  referenceTreeJson (full transitive tree keyed by name, with globalIds and
  child references):
    {
      "com.example.Address": {
        "subject":"address-value","version":1,"globalId":10,
        "schemaType":"AVRO","schema":"{...}",
        "references":[{"name":"com.example.Country","subject":"country-value","version":1}]
      },
      "com.example.Country": {
        "subject":"country-value","version":1,"globalId":20,
        "schemaType":"AVRO","schema":"{...}","references":[]
      }
    }

  The parser joins the two on name and returns the direct refs enriched with
  globalId from the tree. Output for the above:
    [SchemaReferenceEntry(name="com.example.Address", subject="address-value",
                          version=1, globalId=10)]

  Only direct references show up (Country is transitive - it's traversed
  separately by SchemaBackupOrchestrator during backup).

This is needed Needed so restore can register schemas in dependency order the parent schema can't be registered in the destination SR until every direct reference it depends on already exists with a matching ID. Hope this helps.

core/src/main/java/io/confluent/connect/storage/backup/ConverterTypeDetector.java
sribaabu: It was added as an extension hook for custom converters but nothing calls it today and don't think we need it. Removed. Thanks for flagging.

core/src/main/java/io/confluent/connect/storage/backup/ConverterTypeDetector.java
sribaabu: All of them are Kafka Connect's built-in primitive converters. Integer/Long/String/ByteArray are commonly used for keys. Short/Float/Double are rare in practice but valid Connect converters, so we cover them for symmetry.

core/src/main/java/io/confluent/connect/storage/backup/ConverterTypeDetector.java
sribaabu: Thanks @subhashiyer9 Good catch. Backup writes keySchemaType/valueSchemaType into every envelope and uses it to decide whether to back up schema files (only SR-backed types qualify). If the user relied on worker.properties defaults we saw null, tagged UNKNOWN, skipped the SR-backed check, and silently wrote envelopes with no schema files - unrecoverable on restore. Fixed this gap by failing sink startup when either converter isn't set explicitly, and folded the null branch into TYPE_UNKNOWN so we only carry one "unrecognized" tag downstream.

core/src/main/java/io/confluent/connect/storage/StorageSinkConnectorConfig.java
sribaabu: Agreed, Reverted the refactor.

core/src/main/java/io/confluent/connect/storage/backup/EnvelopeSchemaBuilder.java
sribaabu: OPTIONAL_STRING_SCHEMA is the fallback only when the record has no schema
  (keySchema/valueSchema is null). Two things drive the choice:

  - OPTIONAL_ handles nullability - regular records can have null keys
    (round-robin partitioning), and tombstones carry a null value.
  - STRING covers the schemaless path (JsonConverter with schemas.enable=false) - by the time it reaches
    this builder, the data has already been coerced to a JSON string upstream.

Records with a real Connect schema (Avro/Protobuf/JsonSchema/primitive converters) use that schema directly and the fallback never kicks in.

format/src/main/java/io/confluent/connect/storage/format/backup/BackupWrapperExtractor.java
sribaabu: Updated.

format/src/main/java/io/confluent/connect/storage/format/backup/BackupWrapperExtractor.java
sribaabu: Yes - passthrough is the branch that handles primitive converters (StringConverter, IntegerConverter, LongConverter, ByteArrayConverter, ...). These produce data WITH a Connect schema (STRING_SCHEMA, INT32_SCHEMA, ...) but are not SR-backed so they don't wrap in BackupWrapper. 

Falls into passthrough: schema and data are carried into the envelope as-is, and schema-file backup is skipped (no rawSchema for primitives).

Schemaless is a different branch - triggered only when schema == null (JsonConverter with schemas.enable=false, which returns a raw Map/List).

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
sribaabu: It depends on whether the converter is SR-backed:
     - SR-backed converter without the flag: FAILS validation at startup
     - Non-SR converter (StringConverter/IntegerConverter/ByteArrayConverter/etc.):
       flag is not required, validation skips the check entirely.
     So valid setups include: SR-backed on value only, SR-backed on both,
     or SR-backed on neither. The failure case is "SR-backed converter that
     forgot the flag" on either side.

Also fixed the startup log gap you flagged - now shows schema.backup.enabled for both key and value converters instead of just value.

format/src/main/java/io/confluent/connect/storage/format/backup/EnvelopeTransformer.java
sribaabu: **Malformed/invalid records**

  wrap() throws DataException for:
  - Both key AND value are null.
  - Corrupt wrapper struct (missing data field, null wrapper schema)- via BackupWrapperExtractor
  - Schemaless JSON serialization failure - same
  
  All exceptions propagate to put() → Kafka Connect's error handling framework. If user has errors.tolerance=all with a DLQ, the record goes there; otherwise the task fails.

For **behavior.on.null.values**

We wrap before S3SinkTask.put() runs maybeSkipOnNullValue. After wrapping:
  - Every record's value is a KafkaRecordEnvelope Struct (non-null), even for tombstones- the tombstone's null value lives INSIDE the envelope.
  - record.value() == null in maybeSkipOnNullValue is always false.
  - So behavior.on.null.values=IGNORE/FAIL/WRITE has no effect - tombstones are always written.

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
sribaabu: Thanks for flagging, Added validation:
  - Allow: DefaultPartitioner, TimeBasedPartitioner, DailyPartitioner,
    HourlyPartitioner
  - Reject: FieldPartitioner and any custom partitioner class
  - Reject: timestamp.extractor=RecordFieldTimestampExtractor (Wallclock and
    Record extractors are fine)

format/src/main/java/io/confluent/connect/storage/format/backup/SchemaBackupOrchestrator.java
sribaabu: Right, the two checks were doing different things and shouldn't have shared
  a branch. Splitted them

  - globalId <= 0 now throws DataException. SR should never emit an
    invalid globalId, so this is a real corruption signal - failing loud is
    better than silently dropping the reference.
  - The visited-set add stays as a silent return - that's normal cycle
    prevention when a reference tree has cross-links.

format/src/main/java/io/confluent/connect/storage/format/backup/BackupWrapperExtractor.java
sribaabu: SR converters don't have this pattern - they're always schemaful. The schemaless path only kicks in for JsonConverter with schemas.enable=false, which returns Map, List, or a JSON scalar (String, Number, Boolean).

That said, checking the restore side surfaced a real gap: the Map/List  branch went through Jackson, but the else branch used .toString() which worked for Number/Boolean (their toString matches JSON) but dropped the
quotes on String - "hello".toString() = hello, not "hello". Restore does OBJECT_MAPPER.readValue(...) at EnvelopeStorageObjectFormat. so bare hello threw JsonParseException.

Fixed by using Jackson uniformly for all schemaless values. Byte-identical output for Map/List/Number/Boolean/null (only String case actually changes), lossless round-trip, no shape or schema change.

format/src/main/java/io/confluent/connect/storage/format/backup/SchemaBackupOrchestrator.java
```diff
@@ -0,0 +1,185 @@
...
+      String refName = (String) m.get(BackupEnvelope.REF_FIELD_NAME);
+      String refSubject = (String) m.get(BackupEnvelope.REF_FIELD_SUBJECT);
+      int refVersion = m.get(BackupEnvelope.REF_FIELD_VERSION) instanceof Number
+          ? ((Number) m.get(BackupEnvelope.REF_FIELD_VERSION)).intValue() : 0;
+      Map<String, Object> childNode = tree.get(refName);
+      int refGlobalId = childNode != null
+          && childNode.get(BackupEnvelope.REF_FIELD_GLOBAL_ID) instanceof Number
+          ? ((Number) childNode.get(BackupEnvelope.REF_FIELD_GLOBAL_ID)).intValue() : 0;
+      childRefs.add(new SchemaManifest.SchemaReferenceEntry(
+          refName, refSubject, refVersion, refGlobalId));
```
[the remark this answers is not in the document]
sribaabu: Fixed.

pom.xml
```diff
@@ -157,6 +158,12 @@
                 <version>${confluent.version.range}</version>
                 <scope>provided</scope>
             </dependency>
+            <dependency>
+                <groupId>io.confluent</groupId>
+                <artifactId>kafka-connect-schema-backup-api</artifactId>
+                <version>${schema.backup.api.version}</version>
+                <scope>provided</scope>
+            </dependency>
```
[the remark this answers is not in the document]
sribaabu: Fixed.

core/src/main/java/io/confluent/connect/storage/backup/BackupEnvelope.java
subhashiyer9: Is there any use case of this?

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
subhashiyer9: Just curious, what is the reason to not be able to support compression?

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
subhashiyer9: Do customers need to consciously set this config now? Or is it possible to override it specifically for backup/restore?
Header converters are typically not something anyone configures in most cases, and uses default config. Now it becomes mandatory to set it to ByteArrayConverter

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
subhashiyer9: Wondering if there is a way to have single `schema.backup.enabled` that applies to all converters.
Setting it to true on non-SR converter is a failure or no-op?

core/src/main/java/io/confluent/connect/storage/backup/BackupEnvelope.java
sribaabu: No use case yet, but its informational only. Just that we capture timestamp, We also capture its type since it is available. Could be useful for analytical workloads if at all needed or to be interpret the time stamp.

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
sribaabu: image


On the screenshot above,  I noticed the current backup and restore document explicitly mentioned the compression isn't supported and in parquet case by default we enable snappy compression so just want to leave the current behavior as is. Looks like the current implementation on the source doesn't decompress while reading it back.

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
sribaabu: For CP yes, its a connect runtime limitation now, the converters gets instantiated with the converter specific configs in the runtime. Without changing the connect ce runtime code, this dynamism is not possible in CP. All the converters gets initialized picking up either worker or the supplied converter configs. 

For CC, I can try the current templating to dynamically pick or set these.

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
sribaabu: Setting it to true for Non-SR Converter is no-op and doesn't indicate any failures. Should that be ok? 

Also the same limitation of header.converter where all the converters gets its own config with key.converter / value.converter prefix passed to respective converter classes and gets initialized, for CP we don't have other option rather than to send it separately without connect-ce

But for CC, could try with the template override.

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
subhashiyer9: This looks like a docs gap. Source connector does support reading compressed files

subhashiyer9 (approved):

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
```diff
@@ -0,0 +1,482 @@
...
+   */
+  public static List<String> validateSourceConfigs(
+      Map<String, String> configs,
+      String formatClassName,
+      boolean jsonSchemaEmbedded,
+      String modeName) {
+    List<String> errors = new ArrayList<>();
+
+    validateByteArrayFormat(formatClassName, errors);
+    validateJsonFormatSchemaEnable(formatClassName, jsonSchemaEmbedded, modeName, errors);
```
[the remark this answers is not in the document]
sribaabu: Compression Validation is removed and the Parquet compression behavior is fixed.

core/src/main/java/io/confluent/connect/storage/backup/BackupModeValidator.java
sribaabu: Parquet Format has a <compression type>.parquet in the file name convention, which source connector regex skips it and I've fixed it in the source connectors.

core/src/main/java/io/confluent/connect/storage/backup/EnvelopeSchemaBuilder.java
```diff
@@ -0,0 +1,151 @@
...
+    List<Struct> result = new ArrayList<>();
+    if (sinkRecord.headers() != null) {
+      for (Header h : sinkRecord.headers()) {
+        Struct hs = new Struct(HEADER_SCHEMA);
+        hs.put(BackupEnvelope.FIELD_HEADER_KEY, h.key());
+        hs.put(BackupEnvelope.FIELD_HEADER_VALUE,
+            h.value() != null
+                ? Values.convertToString(h.schema(), h.value()) : null);
+        hs.put(BackupEnvelope.FIELD_HEADER_SCHEMA_TYPE,
+            h.schema() != null ? h.schema().type().name() : null);
```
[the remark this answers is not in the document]
sribaabu: Pristine round-trip works when using header.converter=ByteArrayConverter (schema=BYTES, value=byte[], base64 round-trip via BackupHeaderCodec). For other converters, primitives survive but logical/STRUCT/ARRAY/MAP headers lose schema fidelity as noted.
  
 Elevated the guardrail to log.warn and rewrote the message to name the concrete loss modes  instead of the soft "consider using". Fail-fast on non-primitive schemas deferred- the  one-line workaround (ByteArrayConverter) covers all cases losslessly,  and users on SimpleHeaderConverter with primitive-only headers would  be blocked unfairly.

